PR #26, opened 2016-07-22
case insesitve equals & contains

godrei:


stringutil/stringutil_test.go
```diff
@@ -6,6 +6,30 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestCaseInsensitiveEquals(t *testing.T) {
+	require.Equal(t, true, CaseInsensitiveEquals("a", "a"))
```
birmacher: test with nil, "" and nil, nil

PR #33, opened 2016-09-18
cmdex outout if error fix

viktorbenei:
return the output even if there was an error running the command!


viktorbenei:
@godrei related issue - missing `brew install` error log - https://github.com/bitrise-io/steps-amazon-s3-deploy/issues/19

PR #35, opened 2016-09-19
simple timeout package

godrei:


godrei:
Pls add a unit test, to show how this pkg can be used.

Also, https://github.com/bitrise-io/go-utils/blob/master/progress/progress.go#L20 - a "simple" abort mode

viktorbenei:
should this be closed @godrei ?

godrei:
yes, it should.

PR #40, opened 2016-12-08
text with indent and max length

viktorbenei:


stringutil/text.go
```diff
@@ -0,0 +1,49 @@
+package stringutil
+
+import (
+	"bufio"
+	"math"
+	"strings"
+)
+
+// IndentTextWithMaxLength ...
+func IndentTextWithMaxLength(text string, indent string, maxTextLineCharWidth int, isIndentFirstLine bool) string {
```
godrei: instead of `text string, indent string`

write `text, indent string`

stringutil/text.go
```diff
@@ -0,0 +1,49 @@
...
+	if maxTextLineCharWidth < 1 {
+		return ""
+	}
+
+	formattedText := ""
+
+	addLine := func(line string) {
+		isFirstLine := (formattedText == "")
+		if !isFirstLine {
+			formattedText = formattedText + "\n"
```
godrei: why do not use `formattedText += "\n"` ?

stringutil/text.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		if isFirstLine && !isIndentFirstLine {
+			formattedText = line
+		} else {
+			formattedText = formattedText + indent + line
```
godrei: also here `formattedText += indent + line`

stringutil/text.go
```diff
@@ -0,0 +1,49 @@
...
+func IndentTextWithMaxLength(text string, indent string, maxTextLineCharWidth int, isIndentFirstLine bool) string {
+	if maxTextLineCharWidth < 1 {
+		return ""
+	}
+
+	formattedText := ""
+
+	addLine := func(line string) {
+		isFirstLine := (formattedText == "")
+		if !isFirstLine {
```
godrei: i do not really like this if-else block.

i think we should either use:

```
if !isFirstLine {
...
} else {
  if !isIndentFirstLine {
  ...
  } else {
  ...
  }
}
```

or

```
if !isFirstLine {
...
  return
}

if !isIndentFirstLine {
...
} else {
...
}
```

or 

```
if !isFirstLine {
...
} else if !isIndentFirstLine {  
...
} else {
...
}
```

PR #41, opened 2016-12-08
expand log package

godrei:
# colorstring package:

- ColorFunc & ColorfFunc, to allow refer to the coloring functions with type (`printfWithColor(color colorstring.ColorfFunc, format string, v ...interface{})`)
- NoColor & NoColorf, to allow print without color

```
func Printf(format string, v ...interface{}) {
  printfWithColor(colorstring.NoColorf, format, v...)
}
```

# log package:

- formatted logging method names got `f` suffix, to be consistent
- Logger interface, Formatable interface & default Message struct
- JSON and Raw Loggers

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -1,0 +1,37 @@
+package log
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// JSONLoger ...
+type JSONLoger struct {
+	writer io.Writer
+}
+
+// NewJSONLoger ...
+func NewJSONLoger(writer io.Writer) *JSONLoger {
+	return &JSONLoger{
+		writer: writer,
+	}
+}
+
+// NewDefaultJSONLoger ...
+func NewDefaultJSONLoger() JSONLoger {
+	return JSONLoger{
+		writer: os.Stdout,
+	}
+}
+
+// PrintO ...
+func (l JSONLoger) PrintO(f Formatable) {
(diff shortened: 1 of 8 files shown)


log/json_logger.go
```diff
@@ -0,0 +1,37 @@
...
+
+// NewDefaultJSONLoger ...
+func NewDefaultJSONLoger() JSONLoger {
+	return JSONLoger{
+		writer: os.Stdout,
+	}
+}
+
+// PrintO ...
+func (l JSONLoger) PrintO(f Formatable) {
```
viktorbenei: Print0?

log/logger.go
```diff
@@ -0,0 +1,54 @@
...
+// Formatable ...
+type Formatable interface {
+	String() string
+	JSON() string
+}
+
+// Message ...
+type Message struct {
+	Content  interface{} `json:"content,omitempty"`
+	Error    string      `json:"error,omitempty"`
```
viktorbenei: shouldn't this be `errors` - array?

log/logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Formatable ...
+type Formatable interface {
+	String() string
+	JSON() string
+}
+
+// Message ...
+type Message struct {
+	Content  interface{} `json:"content,omitempty"`
```
viktorbenei: I'd go with `Data` instead

PR #58, opened 2017-07-27
ruby script runner package

godrei:


command/rubyscript/rubyscript_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const gemfileContent = `# frozen_string_literal: true
+source "https://rubygems.org"
+
+gem "json"
+`
+
+const gemfileLockContet = `GEM
```
viktorbenei: I think gemfile.lock could / should be optional. You can specify the version directly in Gemfile if required

PR #59, opened 2017-08-04
interface -> map[string]interface casting

godrei:


pointers/pointers_test.go
```diff
@@ -134,6 +134,104 @@ func TestNewInt64Ptr(t *testing.T) {
 	}
 }
 
+/*
```
trapacska: If no any reason to keep uncommented code, either uncomment it, or remove please

PR #71, opened 2017-11-27
Revision command/git package

kdobmayer:


command/git/commands.go
```diff
@@ -0,0 +1,145 @@
+package git
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/bitrise-io/go-utils/command"
+)
+
+// Init creates an empty Git reporitory or reinitializes an existing one.
```
godrei: use `.` or not, but do not mix its usage

command/git/git.go
```diff
@@ -1,209 +1,25 @@
...
+	return &Git{path: path}
 }
 
-// GetCommitHashOfHead ...
-func GetCommitHashOfHead(pth string) (string, error) {
-	cmd := GetCommitHashOfHeadCommand()
-	cmd.SetDir(pth)
-	return cmd.RunAndReturnTrimmedCombinedOutput()
+// Command returns with a Git command.
+func (g *Git) Command(args ...string) *command.Model {
```
godrei: i think this should not be exported.

command/git/commands.go
```diff
@@ -0,0 +1,145 @@
...
+	args := []string{"fetch"}
+	if depth != 0 {
+		args = append(args, "--depth="+strconv.Itoa(depth))
+	}
+	args = append(args, opts...)
+	return g.Command(args...)
+}
+
+// FetchPR downloads objects and refs for the pull request branch.
+func (g *Git) FetchPR(branch string) *command.Model {
```
godrei: do we need this in the lib? if yes it is only for github prs, right? if yes rename it!

command/git/commands.go
```diff
@@ -0,0 +1,145 @@
...
+	return g.Command(args...)
+}
+
+// Pull incorporates changes from a remote repository into the current branch.
+func (g *Git) Pull() *command.Model {
+	return g.Command("pull")
+}
+
+// Add file contents to the index
+func (g *Git) Add(pathspec string) *command.Model {
```
godrei: `pathspec` is not it a single file path? if yes rename to to `pth string`

PR #73, opened 2017-12-08
create git dir if not exists

godrei:


command/git/git.go
```diff
@@ -8,17 +8,20 @@ import (
...
 type Git struct {
-	path string
+	dir string
 }
 
 // New creates a new git project.
-func New(path string) *Git {
-	return &Git{path: path}
+func New(dir string) (Git, error) {
+	if err := os.MkdirAll(dir, 0777); err != nil {
```
kdobmayer: permission should be less open (0755)

PR #76, opened 2018-04-24
removed destination input from CloneTagOrBranch

godrei:


command/git/commands.go
```diff
@@ -14,6 +14,11 @@ func (g *Git) Clone(repo string) *command.Model {
 	return g.command("clone", repo, ".")
 }
 
+// CloneTagOrBranch is recursively clones a tag or branch.
+func (g *Git) CloneTagOrBranch(repo, tagOrBranch string) *command.Model {
+	return g.command("git", "clone", "--recursive", "--branch", tagOrBranch, repo, ".")
```
trapacska: `"git"` part of the command is needed/correct here? (the other commands has no git in their command slice)

godrei: awesome catch, thanks!

PR #77, opened 2018-05-16
added logger

trapacska:


log/logger.go
```diff
@@ -10,3 +10,59 @@ type Formatable interface {
...
+
+// DefaultLogger ...
+type DefaultLogger struct {
+	ts bool
+}
+
+type logfunc func(string, ...interface{})
+
+// Timestamp ...
+func (dl DefaultLogger) Timestamp() DefaultLogger {
```
godrei: we may want to enable and later disable the timestamp logging for the same logger instance.

PR #78, opened 2018-11-12
Support new ruby path

timwangdev:
```sh
$PATH=/Users/vagrant/.bitrise/tools:/usr/local/opt/ruby/bin:/usr/local/bin:/usr/local/sbin:~/bin:/usr/local/bin:...
```
Current mac build use `/usr/local/opt/ruby/bin` in `PATH` as resolved ruby path, which will get Unknown install type when running ruby script (like CocoaPods). This patch should fix this.

### DIFF
--- a/command/rubycommand/rubycommand.go
+++ b/command/rubycommand/rubycommand.go
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	systemRubyPth = "/usr/bin/ruby"
-	brewRubyPth   = "/usr/local/bin/ruby"
+	systemRubyPth  = "/usr/bin/ruby"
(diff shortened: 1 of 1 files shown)


viktorbenei:
Thanks for the PR @timwangdev ! Our CLI & Infra teams are looking into the related issue, but your PR seems to be a promising solution! 😉

adebree:
Actually, the error is `Failed to create command, error: unkown ruby installation type` (pasting this for google / search reasons, especially because of the typo in unkown)

viktorbenei:
Thank you guys and sorry for the inconvenience! We also fixed this on an infrastructure level on all affected stacks, so it should work for every step now, but we'll also update the related steps with this fix.

PR #81, opened 2019-01-17
Use Magenta color for Debug log messages.

viktorbenei:
In order to be able to distinguish Debug log messages, I'd propose to use a separate color for it. Tested Magenta in a few scenarios/color schemes and it seemed fitting for Debug.


viktorbenei:
@lszucs when I was testing it, out of the available colors in https://github.com/bitrise-io/go-utils/blob/2a09aab8380d7842750328aebd5671bcccea89c8/colorstring/colorstring.go#L12 magenta seemed like the nest, still unused option, as it pops out from the rest of the log. 

I think there are two things to consider for debug logs:

1. Is what you mentioned that debug logs are usually verbose.
1. On the other hand debug logs should NOT be used in production, and when you use Debug logs you actually really want to know those.

I opted for the 2nd option, to be able to quickly identify debug logs as in the tool i was working in I wanted to make sure debug logs are not enabled in production/only used while debugging as well as when I enabled it I knew that I'm primarily interested in those logs only (as the non debug logs did not include the details I was looking for).

BirmacherAkos:
What happened with this PR?
@viktorbenei @trapacska

viktorbenei:
Merged - @BirmacherAkos it seems everybody approved it but nobody merged it 🤷‍♂

viktorbenei:
Ohh, actually @lszucs had some concerns - @lszucs should I roll this back? Or should we test it and if needed change the color code to something else?

PR #85, opened 2019-03-02
envutil: RequiredEnv - for less error-prone required env var getter.

viktorbenei:
Use this function to reduce error prone code duplication.
E.g. instead of doing this in your code:

```
myVar1 := os.Getenv("MY_ENV1")
if len(myVar1) < 1 {
return nil, errors.New("MY_ENV1 required")
}
```

You can use this function like:

```
myVar1, err := requiredEnv("MY_ENV1")
if err != nil {
return nil, errors.WithStack(err)
}
```

In the first example you have to specify myVar1 and MY_ENV1 two times, which can lead to
issues if you copy paste that code but e.g. forget to change the var name in the `if len(myVar1) < 1` line,
or you might simply forget to change the var key in the error.


envutil/envutil.go
```diff
@@ -41,3 +44,34 @@ func GetenvWithDefault(envKey, defValue string) string {
...
+// myVar1, err := requiredEnv("MY_ENV1")
+// if err != nil {
+// 	return nil, errors.WithStack(err)
+// }
+// ```
+//
+// In the first example you have to specify myVar1 and MY_ENV1 two times, which can lead to
+// issues if you copy paste that code but e.g. forget to change the var name in the `if len(myVar1) < 1` line,
+// or you might simply forget to change the var key in the error.
+func RequiredEnv(envKey string) (string, error) {
```
trapacska: If you change to the fast-exit way of doing this then you can pull the value assignment into the if for example:
```
func RequiredEnv(envKey string) (string, error) {
	if val := os.Getenv(envKey); len(val) > 0 {
		return val, nil
	}
	return "", fmt.Errorf("Required environment variable (%s) not provided", envKey)
}
```

envutil/envutil.go
```diff
@@ -41,3 +44,34 @@ func GetenvWithDefault(envKey, defValue string) string {
 	}
 	return retVal
 }
+
+// RequiredEnv - returns the env's value if specified, otherwise it returns an error that explains the key is required.
```
trapacska: The format of the comment section is not godoc compatible, I think there are missing indentations. Read more here: https://www.ardanlabs.com/blog/2013/06/documenting-go-code-with-godoc.html

Also the problem visually:
screen shot 2019-03-07 at 10 25 17 am

envutil/envutil.go
```diff
@@ -41,3 +44,34 @@ func GetenvWithDefault(envKey, defValue string) string {
...
+// }
+// ```
+//
+// In the first example you have to specify myVar1 and MY_ENV1 two times, which can lead to
+// issues if you copy paste that code but e.g. forget to change the var name in the `if len(myVar1) < 1` line,
+// or you might simply forget to change the var key in the error.
+func RequiredEnv(envKey string) (string, error) {
+	val := os.Getenv(envKey)
+	if len(val) < 1 {
+		return "", fmt.Errorf("Required environment variable (%s) not provided", envKey)
```
godrei: [Error strings should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

viktorbenei:
Resolved all 3 PR comments - thank you guys!

Please have a look again @godrei @trapacska

PR #87, opened 2019-04-02
Internal logging

lszucs:
Internal logging is about sending certain extra information to our analytics server.

This PR simply extends existing `log.Printf`, `log.Infof`, etc... functions to return a log entry object, that can be sent to the analytics server.

API could be better, sending should be the responsibility of a separate service, but as of now, there is no use case for a dedicated service object -- not to mention, that the card description already suggested an API like this.

### DIFF
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -25,25 +25,27 @@ func init() {
 
 // SendToInternal sends the log message to the configured analytics server
 func (lm Message) SendToInternal(stepID, tag string, data map[string]interface{}) {
-	for k, v := range data {
-		lm.Data[k] = v
-	}
-
-	lm.Data["step_id"] = stepID
-	lm.Data["tag"] = tag
-
-	b, err := json.Marshal(lm)
-	if err != nil {
-		fmt.Printf("marshal log message: %s\n", err)
-	}
-
-	resp, err := http.Post(analyticsServerURL + "/logs", "application/json", bytes.NewReader(b))
-	if err != nil {
-		fmt.Printf("post log message: %s\n", err)
-	}
-
-	if resp.StatusCode != 200 {
-		fmt.Printf("post log message response: %s %s\n", resp.Status, resp.Body)
+	if enableDebugLog {
+		for k, v := range data {
+			lm.Data[k] = v
+		}
+	
+		lm.Data["step_id"] = stepID
+		lm.Data["tag"] = tag
+	
+		b, err := json.Marshal(lm)
+		if err != nil {
+			fmt.Printf("marshal log message: %s\n", err)
+		}
+	
+		resp, err := http.Post(analyticsServerURL + "/logs", "application/json", bytes.NewReader(b))
+		if err != nil {
+			fmt.Printf("post log message: %s\n", err)
+		}
+	
+		if resp.StatusCode != 200 {
+			fmt.Printf("post log message response: %s %s\n", resp.Status, resp.Body)
+		}	
 	}
 }
 
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -52,19 +52,16 @@ func (e Entry) Internal(stepID, tag string, data map[string]interface{}) {
 		return
 	}
 
-	ctx, cancel := context.WithCancel(context.TODO())
-	_ = time.AfterFunc(3 * time.Second, func() {
-		cancel()
-	})
-
+	ctx, cancel := context.WithTimeout(context.Background(), 3 * time.Second)
+	defer cancel()
+	
 	req, err := http.NewRequest(http.MethodPost, analyticsServerURL  + "/logs", &b)
 	if err != nil {
 		// deliberately not writing into users log
 		return
 	}
-	
-	req.Header.Add("Content-Type", "application/json")
 	req = req.WithContext(ctx)
+	req.Header.Add("Content-Type", "application/json")
 	
 	if _, err := httpClient.Do(req); err != nil {
 		// deliberately not writing into users log
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -1,0 +1,12 @@
+package log
+
+type logMessage struct{
+	LogLevel string `json:"log_level"`
+	Message string `json:"message"`
+	Data map[string]interface{} `json:"data"`
+}
+
+func (lm logMessage) SendToInternal(stepID, tag string, data map[string]interface{}) {
+	// todo: post to analytics server
+}
+
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -27,6 +27,13 @@ func (lm Message) SendToInternal(stepID, tag string, data map[string]interface{}
 		lm.Data[k] = v
 	}
 
+	if v, ok := lm.Data["step_id"]; ok {
+		fmt.Printf("internal logger: data.step_id (%s) will be overriden with (%s) ", v, stepID)
+	}
+	if v, ok := lm.Data["tag"]; ok {
+		fmt.Printf("internal logger: data.tag (%s) will be overriden with (%s) ", v, tag)
+	}
+
 	lm.Data["step_id"] = stepID
 	lm.Data["tag"] = tag
 
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -10,7 +10,8 @@ import (
 
 var analyticsServerURL = "https://bitrise-step-analytics.herokuapp.com"
 
-type logMessage struct{
+// Message represents a line in a log
+type Message struct{
 	LogLevel string `json:"log_level"`
 	Message string `json:"message"`
 	Data map[string]interface{} `json:"data"`
@@ -22,7 +23,8 @@ func init() {
 	}
 }
 
-func (lm logMessage) SendToInternal(stepID, tag string, data map[string]interface{}) {
+// SendToInternal sends the log message to the configured analytics server
+func (lm Message) SendToInternal(stepID, tag string, data map[string]interface{}) {
 	for k, v := range data {
 		lm.Data[k] = v
 	}
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -15,8 +15,8 @@ var (
 	}
 )
 
-// Message represents a line in a log
-type Message struct{
+// Entry represents a line in a log
+type Entry struct{
 	LogLevel string `json:"log_level"`
 	Message string `json:"message"`
 	Data map[string]interface{} `json:"data"`
@@ -30,7 +30,7 @@ func SetAnalyticsServerURL(url string) {
 }
 
 // Internal sends the log message to the configured analytics server
-func (lm Message) Internal(stepID, tag string, data map[string]interface{}) {
+func (lm Entry) Internal(stepID, tag string, data map[string]interface{}) {
 	lm.Data = make(map[string]interface{})
 	for k, v := range data {
 		lm.Data[k] = v
--- a/log/internal_logger.go
+++ b/log/internal_logger.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


log/print.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-func printf(severity Severity, withTime bool, format string, v ...interface{}) {
+func printf(severity Severity, withTime bool, format string, v ...interface{}) string {
```
lpusok: Function name is confusing at first read as Printf is already known.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Message represents a line in a log
+type Message struct{
+	LogLevel string `json:"log_level"`
+	Message string `json:"message"`
+	Data map[string]interface{} `json:"data"`
+}
+
+func init() {
+	if url := os.Getenv("ANALYTICS_SERVER_URL"); url != "" {
```
lpusok: It is not a good idea to get paramters from os env.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+		return
+	}
+
+	lm.Data = make(map[string]interface{})
+	for k, v := range data {
+		lm.Data[k] = v
+	}
+
+	lm.Data["step_id"] = stepID
+	lm.Data["tag"] = tag
```
lpusok: If can return an error, should not quietly owerwrite keys.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+var analyticsServerURL = "https://bitrise-step-analytics.herokuapp.com"
+
+// Message represents a line in a log
+type Message struct{
+	LogLevel string `json:"log_level"`
+	Message string `json:"message"`
+	Data map[string]interface{} `json:"data"`
+}
+
+func init() {
```
godrei: other places we use Set functions to modify default values/configuration: https://github.com/bitrise-io/go-utils/blob/master/log/log.go#L13
I would go this way as `ANALYTICS_SERVER_URL` may a too common name, probably used by other tools, other than that it would be much clear what changed the analytics server URL.

log/print.go
lszucs: Indeed, however, this function is preexisting and I can't come up with a better name at this point.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	if url := os.Getenv("ANALYTICS_SERVER_URL"); url != "" {
+		analyticsServerURL = url
+	}
+}
+
+// SendToInternal sends the log message to the configured analytics server
+func (lm Message) SendToInternal(stepID, tag string, data map[string]interface{}) {
```
trapacska: `SendToInternal` for me makes no sense. The original `Internal` in the task description would fit better, but that is not a very common name for this type of method. I recommend to find a different name for this.
For example I have a T<Log> and I want that to be saved in our logging tool as well. I would try something like:
- T<Log>.Record()
- T<Log>.Store()
- T<Log>.Send()
- T<Log>.....??()

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	if lm.LogLevel == "" {
```
trapacska: In cases we call this method we want something to be stored anyway, I think the network call should happen even if this information is missing.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+	lm.Data["step_id"] = stepID
+	lm.Data["tag"] = tag
+
+	b, err := json.Marshal(lm)
+	if err != nil {
+		fmt.Printf("marshal log message: %s\n", err)
+	}
+
+	resp, err := http.Post(analyticsServerURL + "/logs", "application/json", bytes.NewReader(b))
```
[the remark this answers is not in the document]
trapacska: yes, timeout is important

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+	lm.Data = make(map[string]interface{})
+	for k, v := range data {
+		lm.Data[k] = v
+	}
+
+	lm.Data["step_id"] = stepID
+	lm.Data["tag"] = tag
+
+	b, err := json.Marshal(lm)
```
trapacska: what about having a bytes.Buffer, that can be set for json.NewEncode(<here>).Encode(lm) and also can be passed to the request body reader as well

```
var b bytes.Buffer
 json.NewEncode(&b).Encode(lm)
http.Post(analyticsServerURL + "/logs", "application/json", &b)
```

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		fmt.Printf("marshal log message: %s\n", err)
```
trapacska: there was an error, here you should return to not to call the network call with wrong data. Also, there should be no any logging here.

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	resp, err := http.Post(analyticsServerURL + "/logs", "application/json", bytes.NewReader(b))
+	if err != nil {
+		fmt.Printf("post log message: %s\n", err)
```
trapacska: no return

trapacska: should not log

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if resp.StatusCode != 200 {
+		fmt.Printf("post log message response: %s %s\n", resp.Status, resp.Body)
```
trapacska: ok, return is not required but still should not log

log/internal_logger.go
trapacska: Track I think should be fine, but please check other logging service what they name their method to store a log

log/print.go
```diff
@@ -14,78 +14,126 @@ func printf(severity Severity, withTime bool, format string, v ...interface{}) {
 	if _, err := fmt.Fprintln(outWriter, message); err != nil {
 		fmt.Printf("failed to print message: %s, error: %s\n", message, err)
 	}
+
+	return message
```
trapacska: please return the raw uncolored string here

log/internal_logger.go
```diff
@@ -0,0 +1,54 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"net/http"
+)
+
+var analyticsServerURL = "https://bitrise-step-analytics.herokuapp.com"
+
+// Message represents a line in a log
+type Message struct{
```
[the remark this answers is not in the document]
lszucs: Yeah, nah, been there, done that -- it would be `log.Log` and the linter don't like stuttering. :(

log/internal_logger.go
lszucs: I don't see why `Send` and `Internal` would be fine, but `SendToInternal` make no sense.

I'll rename it to `Record`.

log/internal_logger.go
lpusok: resolved

log/internal_logger.go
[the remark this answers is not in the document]
godrei: What about `Entry` just like in case of logrus: https://github.com/sirupsen/logrus/blob/master/entry.go#L51

log/internal_logger.go
```diff
@@ -0,0 +1,56 @@
...
+	LogLevel string `json:"log_level"`
+	Message string `json:"message"`
+	Data map[string]interface{} `json:"data"`
+}
+
+// SetAnalyticsServerURL updates the the analytics server collecting the
+// logs. Default value is 'https://bitrise-step-analytics.herokuapp.com'
```
godrei: No need for this comment: `Default value is 'https://bitrise-step-analytics.herokuapp.com'`

log/internal_logger.go
```diff
@@ -0,0 +1,56 @@
...
+	lm.Data = make(map[string]interface{})
+	for k, v := range data {
+		lm.Data[k] = v
+	}
+
+	if v, ok := lm.Data["step_id"]; ok {
+		fmt.Printf("internal logger: data.step_id (%s) will be overriden with (%s) ", v, stepID)
+	}
+	if v, ok := lm.Data["tag"]; ok {
+		fmt.Printf("internal logger: data.tag (%s) will be overriden with (%s) ", v, tag)
```
godrei: pls do not print this issues (it would be strange in the build logs and would not make any sense for the users)

log/internal_logger.go
```diff
@@ -0,0 +1,56 @@
...
+
+	lm.Data["step_id"] = stepID
+	lm.Data["tag"] = tag
+
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(lm); err != nil {
+		return
+	}
+
+	_, err := netClient.Post(analyticsServerURL + "/logs", "application/json", &b)
```
godrei: NOTE: did not you like my idea about sending the log asynchronous?
anyway pls move this inside the if block

log/print.go
```diff
@@ -14,78 +14,126 @@ func printf(severity Severity, withTime bool, format string, v ...interface{}) {
...
+func Debugf(format string, v ...interface{}) Message {
 	if enableDebugLog {
-		printf(debugSeverity, false, format, v...)
+		return Message{
+			LogLevel: "info",
+			Message: printf(debugSeverity, false, format, v...),
+		}
+	}
+
+	return Message{
```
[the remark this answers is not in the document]
lpusok: resolved

log/internal_logger.go
lszucs: Ouch -- I forgot about it. Fixing right away.

trapacska: In case this will be async, need to take care about finishing async tasks when the process finishes. So most likely sending an internal log in the last couple of lines of main, it won't be sent as the process finishes earlier than the async task.

lszucs: As discussed on Friday, we are not interested in handling this case now, as it is not a critical issue.

log/internal_logger.go
```diff
@@ -0,0 +1,59 @@
...
+
+	lm.Data["step_id"] = stepID
+	lm.Data["tag"] = tag
+
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(lm); err != nil {
+		return
+	}
+
+	go func() {
```
godrei: pls make the requests synchronous but use 5 s timeout, using context.Cancel: use Contexts to cancel a request from: https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

log/internal_logger.go
```diff
@@ -30,30 +31,45 @@ func SetAnalyticsServerURL(url string) {
...
-	if err := json.NewEncoder(&b).Encode(lm); err != nil {
+	if err := json.NewEncoder(&b).Encode(e); err != nil {
 		return
 	}
 
-	go func() {
-		if _, err := netClient.Post(analyticsServerURL + "/logs", "application/json", &b); err != nil {}
-	}()
+	ctx, cancel := context.WithCancel(context.TODO())
+	_ = time.AfterFunc(3 * time.Second, func() {
```
trapacska: You need to close the timer as soon as you know that it won't need to be fired. Hopefully this won't cause any error, but what if the request is already done and a left over timer fires and calls `cancel` that might already will result in a null pointer exception? Not sure how it works, but in case you no need to fire the timer anymore, please stop and destroy it.

godrei: what about this:

```
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel() // Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete:

	req, err := http.NewRequest(http.MethodPost, analyticsServerURL+"/logs", &b)
	if err != nil {
		// deliberately not writing into users log
		return
	}

	req = req.WithContext(ctx)

	req.Header.Add("Content-Type", "application/json")

	if _, err := httpClient.Do(req); err != nil {
		// deliberately not writing into users log
		return
	}
```

PR #88, opened 2019-04-10
Added funcs.go to templateutil to collect useful functions for Go templates

viktorbenei:
Commonly used functions for templates; functions that you can use as `template.FuncMap` functions.

Copied from https://github.com/bitrise-io/gotgen/blob/125332508853e1117c5e8dae9096fa69f0801e7d/cmd/generate.go#L106

I think it'd make sense to collect these functions here, in `go-utils`, so that these can then be used in projects which use `go-utils/templateutil`.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,7 +1,6 @@
-format_version: 5
+format_version: "5"
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
-
-
+project_type: ""
 workflows:
   test:
     steps:
@@ -15,7 +14,19 @@ workflows:
             go get -u gopkg.in/yaml.v2
     - go-list:
         inputs:
-        - exclude: "*/vendor/*\n*/pkcs12*"
-    - golint:
-    - errcheck:
-    - go-test:
+        - exclude: |-
+            */vendor/*
+            */pkcs12*
+    - script@1.1.5:
+        title: Debug print $BITRISE_GO_PACKAGES
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            set -e
+
+            echo "=== BITRISE_GO_PACKAGES ==="
+            echo "$BITRISE_GO_PACKAGES"
+            echo "==========================="
+    - golint: {}
+    - errcheck: {}
+    - go-test: {}
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,4 +1,4 @@
-format_version: "5"
+format_version: "7"
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
 project_type: ""
 workflows:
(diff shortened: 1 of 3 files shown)


viktorbenei:
Strange.. If I run `errcheck "-asserts=true" "-blank=true" "-verbose" "github.com/bitrise-io/go-utils/templateutil"` in any way I can't reproduce this issue..

BirmacherAkos:
If you remove `the github.com/pkg/errors` package from your local computer it will fail for you too.
I've added it to the dep list, it should be fixed now.

viktorbenei:
Wow that's a fantastic catch, thanks @BirmacherAkos !

templateutil/funcs_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/bitrise-io/go-utils/envutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetenv(t *testing.T) {
+	testEnvKey := "KEY_TestGetenv"
+
+	// no env set yet, empty value should be returned
+	require.Equal(t, "", Getenv(testEnvKey))
```
godrei: pls do not use any testing framework for new tests, we want to deintroduce them.

bitrise.yml
```diff
@@ -1,7 +1,6 @@
-format_version: 5
+format_version: "5"
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
-
-
+project_type: ""
 workflows:
   test:
     steps:
```
BirmacherAkos: Could you please update the go-tests for the newer solution
Example:
https://github.com/bitrise-io/go-xcode/blob/master/bitrise.yml#L17-L20

viktorbenei:
To answer the question 

> other than that, why do we duplicate these functions across envutil and templateutil packages, related pr: #85?

I didn't want to base this PR on the other one. But I agree that once that's merged it'd make sense to use those functions.

bitrise.yml
viktorbenei: I checked it but the same steps are used here too, the only difference is that in `go-list` a few `exclude`s are defined here.

viktorbenei:
@BirmacherAkos updated + commented ( https://github.com/bitrise-io/go-utils/pull/88#discussion_r275404049 )

viktorbenei:
I'll close this PR as in its current form I agree it doesn't really add anything new/useful. I might have some more time in the future to add a few more functions and some template function related utility methods from `gotgetn` e.g. the arithmetic functions https://github.com/bitrise-io/gotgen/blob/e4b5c30d4fe7f6f04c98c21d3d0aa71540ea0962/cmd/generate.go#L138 as well as the "Template functions constructor" func https://github.com/bitrise-io/gotgen/blob/e4b5c30d4fe7f6f04c98c21d3d0aa71540ea0962/cmd/generate.go#L105

bitrise.yml
BirmacherAkos: Yeah, you are right. The exclude list misled me for some reason

PR #90, opened 2019-06-20
Adding bundler related commands

lpusok:


command/gems/bundler.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/bitrise-io/go-utils/command"
+	"github.com/bitrise-io/go-utils/command/rubycommand"
+)
+
+// InstallBundlerCommand returns a command to install a specific bundler version
+func InstallBundlerCommand(gemfileLockVersion Version) (*command.Model, error) {
+	installBundlerCmdParams := []string{"gem", "install", "bundler", "--force", "--no-document"}
+	if gemfileLockVersion.Found {
+		installBundlerCmdParams = append(installBundlerCmdParams, []string{"-v", gemfileLockVersion.Version}...)
```
godrei: NOTE: previously you used long flags (`--force`), but here short (`-v`)

command/gems/bundler.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return command.NewFromSlice(installBundlerCmdParams)
```
godrei: NOTE: you could spare the error return param

```
	args := []string{}
	if true {
		args = append(args, "")
	}
	command.New("gem", args...)
```

command/gems/bundler.go
```diff
@@ -0,0 +1,45 @@
...
+	bundleExec := []string{"bundle"}
+	if bundlerVersion.Found {
+		bundleExec = append(bundleExec, "_"+bundlerVersion.Version+"_")
+	}
+	return append(bundleExec, "exec")
+}
+
+// RbenvVersionsCommand retruns a command to print used and available ruby versions if rbenv is installed
+func RbenvVersionsCommand() (*command.Model, error) {
+	if _, err := command.New("which", "rbenv").RunAndReturnTrimmedCombinedOutput(); err != nil {
```
godrei: exec.LookPath could be used, this way you could remove the error return value

command/gems/gemfile_lock.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+)
+
+// Version contains bundler or fastlane version
+type Version struct {
+	Version string
+	Found   bool
+}
+
+// ParseVersionFromBundle  returns the specified grm version parsed from a Gemfile.lock on a best effort basis, for logging purposes only.
```
godrei: typo: grm

command/gems/gemfile_lock.go
```diff
@@ -0,0 +1,89 @@
...
+//     atomos (0.1.3)
+//     babosa (1.0.2)
+//     badge (0.8.5)
+//       curb (~> 0.9)
+//       fastimage (>= 1.6)
+//       fastlane (>= 2.0)
+//       mini_magick (>= 4.5)
+//     claide (1.0.2)
+func ParseVersionFromBundle(gemName string, gemfileLockContent string) (gemVersion Version, err error) {
+	relevantLines := []string{}
```
godrei: var relevantLines []string

command/gems/gemfile_lock.go
```diff
@@ -0,0 +1,89 @@
...
+	lines := strings.Split(gemfileLockContent, "\n")
+
+	specsStart := false
+	for _, line := range lines {
+		trimmed := strings.Trim(line, " ")
+		if trimmed == "" {
```
godrei: NOTE:

```
if strings.Trim(line, " ") == ""
```

command/gems/gemfile_lock.go
```diff
@@ -0,0 +1,89 @@
+package gems
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// Version contains bundler or fastlane version
```
godrei: or any gem version, right?

command/gems/gemfile_lock.go
```diff
@@ -0,0 +1,89 @@
...
+			continue
+		}
+
+		if specsStart {
+			relevantLines = append(relevantLines, line)
+		}
+	}
+
+	//     fastlane (1.109.0)
+	exp := regexp.MustCompile(fmt.Sprintf(`^%s \((.+)\)`, regexp.QuoteMeta(gemName)))
```
godrei: NOTE: there should be a way to explicitly define that the first `)` should form the match to avoid having issues with lines like: `fastlane (1.109.0) # could be bumped to fastlane (1.110.0)`

command/gems/gemfile_lock_test.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+	tests := []struct {
+		name               string
+		gemName            string
+		gemfileLockContent string
+		wantGemVersion     Version
+		wantErr            bool
+	}{
+		{
+			gemfileLockContent: gemfileLockContent,
```
godrei: forgot to define the test names

command/gems/gemfile_lock.go
lpusok: It should in this case, because ^ is matching the beginning of text or line.

godrei: very nice

PR #92, opened 2019-06-29
Add command to command/git for setting a config value for the repository

arifken:
This PR adds a command to `command/git/commands.go` to allow a `git config` setting to be applied for the current repository.


godrei:
@arifken thanks for the PR!

PR #95, opened 2019-07-18
TOOL-975 Added util to get file name from path.

richard-bogdan-bitrise:


pathutil/pathutil.go
```diff
@@ -179,3 +179,10 @@ func NormalizedOSTempDirPath(tmpDirNamePrefix string) (retPth string, err error)
 	}
 	return
 }
+
+// GetFileName returns the name of the file from a given path or the name of the directory if it is a directory
+func GetFileName(path string) string {
+	file := filepath.Base(path)
```
BirmacherAkos: Note:
It's perfectly fine, but you can get the same result with:
```
{
    return strings.TrimRight(filepath.Base(path), filepath.Ext(path))
}
```

richard-bogdan-bitrise: TrimRight removes the last char of the filename as well, changed to TrimSuffix.

PR #96, opened 2019-07-24
Added RevokableSetenvs and SetenvsForFunction for multi env var handling.

viktorbenei:
Both work the same way as their singular version but instead of accepting a single env var key & value pair these accept a map of key value pairs and set & revoke all of them.

### DIFF
--- a/envutil/envutil.go
+++ b/envutil/envutil.go
@@ -19,10 +19,6 @@ func RevokableSetenv(envKey, envValue string) (func() error, error) {
 func RevokableSetenvs(envs map[string]string) (func() error, error) {
 	origValues := map[string]string{}
 
-	for k := range envs {
-		origValues[k] = os.Getenv(k)
-	}
-
 	revokeFn := func() error {
 		for k, v := range origValues {
 			if err := os.Setenv(k, v); err != nil {
@@ -33,6 +29,8 @@ func RevokableSetenvs(envs map[string]string) (func() error, error) {
 	}
 
 	for k, v := range envs {
+		origValues[k] = os.Getenv(k)
+
 		if err := os.Setenv(k, v); err != nil {
 			return revokeFn, err
 		}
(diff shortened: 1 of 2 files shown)


envutil/envutil.go
```diff
@@ -15,6 +15,32 @@ func RevokableSetenv(envKey, envValue string) (func() error, error) {
 	return revokeFn, os.Setenv(envKey, envValue)
 }
 
+// RevokableSetenvs ...
+func RevokableSetenvs(envs map[string]string) (func() error, error) {
```
godrei: NOTE: with a slight modification you can solve to store and restore only the envs, which was set without error (it also takes less iteration over envs):

```
	origValues := map[string]string{}

	revokeFn := func() error {
		for k, v := range origValues {
			if err := os.Setenv(k, v); err != nil {
				return err
			}
		}
		return nil
	}

	for k, v := range envs {
		origValue := os.Getenv(k)

		if err := os.Setenv(k, v); err != nil {
			return revokeFn, err
		}

		origValues[k] = origValue
	}

	return revokeFn, nil
```

viktorbenei:
@godrei I agree with your suggestion, actually I did one more slight optimisation to make the code shorter

Instead of

```go
	for k, v := range envs {
		origValue := os.Getenv(k)

		if err := os.Setenv(k, v); err != nil {
			return revokeFn, err
		}

		origValues[k] = origValue
	}
```

I used

```go
	for k, v := range envs {
		origValues[k] = os.Getenv(k)

		if err := os.Setenv(k, v); err != nil {
			return revokeFn, err
		}
	}
```

I was thinking about this a bit but I think assigning `origValues[k] = os.Getenv(k)` directly instead of first assigning it to `origValue := os.Getenv(k)` should work just fine, worst case it'll just re-apply the same value when `revokeFn` is called, and the original value should cause no issues (`os.Setenv` should be able to set it, as it already was set).

PR #97, opened 2019-10-15
error message fix

BirmacherAkos:
Provide the error message too not just the error status code.


richard-bogdan-bitrise:
@trapacska pls merge.

PR #98, opened 2019-10-15
Check if the selected ruby version installed via rbenv

BirmacherAkos:
- add a function which returns if the selected ruby version is installed via rbenv or not

### DIFF
--- a/command/rubycommand/rubycommand.go
+++ b/command/rubycommand/rubycommand.go
@@ -209,7 +209,7 @@ func IsGemInstalled(gem, version string) (bool, error) {
 
 	out, err := cmd.RunAndReturnTrimmedCombinedOutput()
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("%s: error: %s", out, err)
 	}
 
 	return findGemInList(out, gem, version)
--- a/command/rubycommand/rubycommand.go
+++ b/command/rubycommand/rubycommand.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/bitrise-io/go-utils/command"
-	"github.com/bitrise-io/go-utils/log"
 	"github.com/bitrise-io/go-utils/pathutil"
 )
 
@@ -216,6 +215,36 @@ func IsGemInstalled(gem, version string) (bool, error) {
 	return findGemInList(out, gem, version)
 }
 
+func isSelectedRbenvRubyInstalled(message string) (bool, string, error) {
+	//
+	// Not installed
+	reg, err := regexp.Compile("rbenv: version \x60.*' is not installed")
+	if err != nil {
+		return false, "", fmt.Errorf("failed to parse regex ( %s ) on the error message, error: %s", "rbenv: version \x60.*' is not installed", err)
+	}
+
+	var version string
+	if reg.MatchString(message) {
+		message := reg.FindString(message)
+		version = strings.Split(strings.Split(message, "`")[1], "'")[0]
+		return false, version, nil
+	}
+
+	//
+	// Installed
+	reg, err = regexp.Compile(".* \\(set by")
+	if err != nil {
+		return false, "", fmt.Errorf("failed to parse regex ( %s ) on the error message, error: %s", ".* \\(set by", err)
+	}
+
+	if reg.MatchString(message) {
+		s := reg.FindString(message)
+		version = strings.Split(s, " (set by")[0]
+		return true, version, nil
+	}
+	return false, version, nil
+}
+
 // IsSelectedRbenvRubyInstalled checks if the selected ruby version is installed via rbenv.
 // Ruby version is set by
 // 1. The RBENV_VERSION environment variable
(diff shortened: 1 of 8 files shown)


command/rubycommand/rubycommand.go
```diff
@@ -213,3 +215,57 @@ func IsGemInstalled(gem, version string) (bool, error) {
...
+// 1. The RBENV_VERSION environment variable
+// 2. The first .ruby-version file found by searching the directory of the script you are executing and each of its
+// parent directories until reaching the root of your filesystem.
+// 3.The first .ruby-version file found by searching the current working directory and each of its parent directories
+// until reaching the root of your filesystem.
+// 4. The global ~/.rbenv/version file. You can modify this file using the rbenv global command.
+func IsSelectedRbenvRubyInstalled(workdir string) (bool, string, error) {
+	absWorkdir, err := pathutil.AbsPath(workdir)
+	if err != nil {
+		return false, "", fmt.Errorf("Failed to expand (%s), error: %s", workdir, err)
```
lszucs: Error message specific to one possible underlying error.

The returned error should clarify the reason for the failure (e.g. problem during expansion) and the outer should only refer to the purpose of the call itself (e.g. `could not get absolute path for workfir`)

Also, please use lowercase initial letter, because you can't be sure if this error will be the topmost error message in the calling code.

Same goes for all error messages below.

command/rubycommand/rubycommand.go
```diff
@@ -213,3 +215,56 @@ func IsGemInstalled(gem, version string) (bool, error) {
...
+// until reaching the root of your filesystem.
+// 4. The global ~/.rbenv/version file. You can modify this file using the rbenv global command.
+func IsSelectedRbenvRubyInstalled(workdir string) (bool, string, error) {
+	absWorkdir, err := pathutil.AbsPath(workdir)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to get absolute path for ( %s ), error: %s", workdir, err)
+	}
+
+	cmd := command.New("rbenv", "version").SetDir(absWorkdir)
+	log.Debugf("$ %s", cmd.PrintableCommandArgs())
```
trapacska: Please do not use log in packages.

command/rubycommand/rubycommand.go
```diff
@@ -213,3 +215,56 @@ func IsGemInstalled(gem, version string) (bool, error) {
...
+
+// IsSelectedRbenvRubyInstalled checks if the selected ruby version is installed via rbenv.
+// Ruby version is set by
+// 1. The RBENV_VERSION environment variable
+// 2. The first .ruby-version file found by searching the directory of the script you are executing and each of its
+// parent directories until reaching the root of your filesystem.
+// 3.The first .ruby-version file found by searching the current working directory and each of its parent directories
+// until reaching the root of your filesystem.
+// 4. The global ~/.rbenv/version file. You can modify this file using the rbenv global command.
+func IsSelectedRbenvRubyInstalled(workdir string) (bool, string, error) {
```
trapacska: pls write unit tests for this

command/rubycommand/rubycommand.go
```diff
@@ -213,3 +215,56 @@ func IsGemInstalled(gem, version string) (bool, error) {
...
+	log.Debugf("Ruby version:")
+	out, err := cmd.RunAndReturnTrimmedCombinedOutput()
+	if err != nil {
+		return false, "", fmt.Errorf("failed to check installed ruby version, error: %s", err)
+	}
+	log.Debugf(out)
+
+	//
+	// Not installed
+	reg, err := regexp.Compile("rbenv: version \x60.*' is not installed")
```
trapacska: this \x60 is very misleading this way, please comment on this

command/rubycommand/rubycommand.go
BirmacherAkos: Not even debug logs? 🤔

command/rubycommand/rubycommand.go
BirmacherAkos: Fixed

command/rubycommand/rubycommand.go
BirmacherAkos: Fixed

command/rubycommand/rubycommand.go
```diff
@@ -213,3 +214,55 @@ func IsGemInstalled(gem, version string) (bool, error) {
...
+
+// IsSelectedRbenvRubyInstalled checks if the selected ruby version is installed via rbenv.
+// Ruby version is set by
+// 1. The RBENV_VERSION environment variable
+// 2. The first .ruby-version file found by searching the directory of the script you are executing and each of its
+// parent directories until reaching the root of your filesystem.
+// 3.The first .ruby-version file found by searching the current working directory and each of its parent directories
+// until reaching the root of your filesystem.
+// 4. The global ~/.rbenv/version file. You can modify this file using the rbenv global command.
+func IsSelectedRbenvRubyInstalled(workdir string) (bool, string, error) {
```
trapacska: NOTE: For me it is very strange that it returns `installed` for the version value given as an input param, however returns a version number that is currently installed.

I think that a logic like:
- getting the currently installed one
- getting a version from the ruby-version file
- comparing them in a main thread

might be more clean and understandable, WDYT?

BirmacherAkos: I removed the version from the return parameters as we don't need it, but I don't want to check the files/ envs, etc manually. Just let the Rbenv to do it.

PR #99, opened 2019-10-24
TOOL-1217

richard-bogdan-bitrise:
Updated removeFile and removeDir to have variadic args.

### DIFF
--- a/command/file.go
+++ b/command/file.go
@@ -33,6 +33,7 @@ func CopyDir(src, dst string, isOnlyContent bool) error {
 }
 
 // RemoveDir ...
+// Deprecated: use RemoveAll instead.
 func RemoveDir(dirPth string) error {
 	if exist, err := pathutil.IsPathExists(dirPth); err != nil {
 		return err
@@ -45,11 +46,22 @@ func RemoveDir(dirPth string) error {
 }
 
 // RemoveFile ...
+// Deprecated: use RemoveAll instead.
 func RemoveFile(pth string) error {
 	if exist, err := pathutil.IsPathExists(pth); err != nil {
 		return err
 	} else if exist {
 		if err := os.Remove(pth); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// RemoveAll removes recursively every file on the given paths.
+func RemoveAll(pths ...string) error {
+	for _, pth := range pths {
+		if err := os.RemoveAll(pth); err != nil {
 			return err
 		}
 	}


richard-bogdan-bitrise:
@trapacska pls merge.

command/file.go
```diff
@@ -33,24 +33,28 @@ func CopyDir(src, dst string, isOnlyContent bool) error {
...
 
 // RemoveDir ...
-func RemoveDir(dirPth string) error {
-	if exist, err := pathutil.IsPathExists(dirPth); err != nil {
-		return err
-	} else if exist {
-		if err := os.RemoveAll(dirPth); err != nil {
+func RemoveDir(dirPths ...string) error {
+	for _, dirPth := range dirPths {
+		if exist, err := pathutil.IsPathExists(dirPth); err != nil {
```
trapacska: pls remove this check, as in case of "remove" the main goal is to not to exists the dirs, and if it is already not exists then RemoveAll will not fail at all or if it is existing the it will remove it. unnecessary step here.

https://golang.org/pkg/os/#RemoveAll

command/file.go
```diff
@@ -33,24 +33,28 @@ func CopyDir(src, dst string, isOnlyContent bool) error {
 }
 
 // RemoveDir ...
-func RemoveDir(dirPth string) error {
-	if exist, err := pathutil.IsPathExists(dirPth); err != nil {
-		return err
-	} else if exist {
-		if err := os.RemoveAll(dirPth); err != nil {
+func RemoveDir(dirPths ...string) error {
```
trapacska: RemoveDir name is not matching for the function logic

command/file.go
```diff
@@ -33,24 +33,28 @@ func CopyDir(src, dst string, isOnlyContent bool) error {
...
+	for _, dirPth := range dirPths {
+		if exist, err := pathutil.IsPathExists(dirPth); err != nil {
 			return err
+		} else if exist {
+			if err := os.RemoveAll(dirPth); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
```
trapacska: except the 'for' it is not necessary to have a single os.RemoveAll command call as it is not an optimization

command/file.go
```diff
@@ -33,24 +33,28 @@ func CopyDir(src, dst string, isOnlyContent bool) error {
...
 }
 
 // RemoveFile ...
-func RemoveFile(pth string) error {
-	if exist, err := pathutil.IsPathExists(pth); err != nil {
-		return err
-	} else if exist {
-		if err := os.Remove(pth); err != nil {
+func RemoveFile(pths ...string) error {
```
trapacska: please use os.RemoveAll instead, to have an exact controlled behaviour (with Remove if you read the comment it is not exact what and how will happen) and also this function should not fail if there is something not existing as it will remove anyway

trapacska: if you will have the same logic for both function then please use shared codebase

trapacska: this function like:
```
func removeAll(pths ...string) error {
	for _, pth := range pths {
		if err := os.RemoveAll(pth); err != nil {
			return err
		}
	}
	return nil
}
```

is not used on so many places, why is it required to move out to a package?

richard-bogdan-bitrise:
Basically I just wanted to let the dev call these methods with variadic args. Due to your comments I decided to deprecate them and create the RemoveAll.

command/file.go
richard-bogdan-bitrise: This is a utility method, which should be available to be used in every package.

PR #103, opened 2019-11-06
TOOL-30 Updated gems related methods to check for gems.locked file too.

richard-bogdan-bitrise:


command/gems/gemfile_lock.go
```diff
@@ -86,3 +89,41 @@ func ParseBundlerVersion(gemfileLockContent string) (gemVersion Version, err err
...
 	}, nil
 }
+
+var gemFileLockNames = []string{
+	"Gemfile.lock", "gems.locked",
+}
+
+// GemFileLockPth gets the path for the gem lock file from the given directory.
+func GemFileLockPth(searchDir string) (string, error) {
+	var pth string
```
lpusok: Would use a setup like this:
```
	for i, gemFileName := range gemFileLockNames {
		pth := filepath.Join(searchDir, gemFileName)
		if _, err := os.Stat(pth); err != nil {
			if os.IsNotExist(err) {
				continue
			}
			return "", err
		}
		return pth, nil
	}
	return "", errors.New("no lock file found")
```

command/gems/gemfile_lock.go
```diff
@@ -86,3 +89,41 @@ func ParseBundlerVersion(gemfileLockContent string) (gemVersion Version, err err
...
+			return "", err
+		} else if err == nil {
+			return pth, nil
+		}
+	}
+	return pth, nil
+}
+
+// GemFileLockContent gets the content of the gem lock file from the given directory.
+func GemFileLockContent(searchDir string) (string, error) {
```
lpusok: Not entirely convinced that GemFileLockContent and ParseVersionFromBundlePth are helpful enough to be added as a package level exported functions.

richard-bogdan-bitrise: Would be needed for fastlane step.

command/gems/gemfile_lock.go
richard-bogdan-bitrise: Updated code.

richard-bogdan-bitrise:
@godrei pls merge.

PR #105, opened 2020-01-05
New fileutil function: WriteJSONToFile

viktorbenei:
Added WriteJSONToFile with tests.

Also added a test for WriteStringToFile, and initialized go modules
for the project.


fileutil/fileutil.go
```diff
@@ -57,6 +58,15 @@ func WriteBytesToFile(pth string, fileCont []byte) error {
 	return WriteBytesToFileWithPermission(pth, fileCont, 0)
 }
 
+// WriteJSONToFile ...
+func WriteJSONToFile(pth string, fileCont interface{}) error {
+	bytes, err := json.Marshal(fileCont)
+	if err != nil {
+		return fmt.Errorf("failed to JSON marshal the provided object: %+v", err)
```
koral--: Shouldn't this case also be tested?

viktorbenei: You're absolutely right, added failure/error tests too.

PR #109, opened 2020-06-29
Adding GemInstallPrerelease function, to install the latest prereleas…

lpusok:
…e version of a gem.

### DIFF
--- a/command/rubycommand/rubycommand.go
+++ b/command/rubycommand/rubycommand.go
@@ -155,9 +155,9 @@ func GemUpdate(gem string) ([]*command.Model, error) {
 func gemInstallCommand(gem, version string, enablePrerelease bool) []string {
 	slice := []string{"gem", "install", gem, "--no-document"}
 	if enablePrerelease {
-		slice = append(slice, "--pre")
+		slice = append(slice, "--prerelease")
 	}
-	if !enablePrerelease && version != "" {
+	if version != "" {
 		slice = append(slice, "-v", version)
 	}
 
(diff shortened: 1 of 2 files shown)


command/rubycommand/rubycommand.go
```diff
@@ -152,21 +152,48 @@ func GemUpdate(gem string) ([]*command.Model, error) {
 	return cmds, nil
 }
 
+func gemInstallCommand(gem, version string, enablePrerelease bool) []string {
+	slice := []string{"gem", "install", gem, "--no-document"}
+	if enablePrerelease {
+		slice = append(slice, "--pre")
```
godrei, line 158: based on the [gem install docs](https://guides.rubygems.org/command-reference/#gem-install), it should be:

```
-​-[no-]prerelease - Allow prerelease versions of a gem to be installed. (Only for listed gems)
```

command/rubycommand/rubycommand.go
```diff
@@ -152,21 +152,48 @@ func GemUpdate(gem string) ([]*command.Model, error) {
 	return cmds, nil
 }
 
+func gemInstallCommand(gem, version string, enablePrerelease bool) []string {
```
godrei, line 155: NOTE: it is a bit strange with this function that either `version` or `enablePrerelease` can be used, but not both.

lpusok: Did not consider that version range can be used with prerelease option.

command/rubycommand/rubycommand.go
lpusok: It worked fine, but updated it.

PR #112, opened 2020-12-07
Improve `RunCmdAndReturnExitCode`

godrei:
**`command/RunCmdAndReturnExitCode` changes:**

`errorutil.CmdExitCodeFromError(err)` does the same as `cmd.ProcessState.ExitCode()`, except CmdExitCodeFromError falls back to exist status 0 or 1 in case of an issue, while Go standard library uses `-1`.

more info:
- https://golang.org/src/os/exec/exec.go?s=13923:13957#L467
- https://golang.org/src/os/exec_posix.go?s=2421:2459#L87

~~**`errorutil` deprecated:**~~

~~All functionalities can be solved by the standard library functions, see deprecation notes and new tests.~~

**`errorutil` changes:**

Use Go standard library functionalities.

### DIFF
--- a/command/command_test.go
+++ b/command/command_test.go
@@ -1,28 +1,30 @@
-package command
+package command_test
 
 import (
+	"os/exec"
 	"testing"
 
+	"github.com/bitrise-io/go-utils/command"
 	"github.com/stretchr/testify/require"
 )
 
 func TestNewCommandSlice(t *testing.T) {
 	t.Log("it fails if slice empty")
 	{
-		cmd, err := NewFromSlice([]string{})
+		cmd, err := command.NewFromSlice([]string{})
 		require.Error(t, err)
-		require.Equal(t, (*Model)(nil), cmd)
+		require.Equal(t, (*command.Model)(nil), cmd)
 	}
 
 	t.Log("it creates cmd if cmdSlice has 1 element")
 	{
-		_, err := NewFromSlice([]string{"ls"})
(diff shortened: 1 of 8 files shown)


errorutil/errorutil.go
```diff
@@ -8,22 +11,30 @@ import (
...
 }
 
 // CmdExitCodeFromError ...
+//
+// Deprecated: use exec.Cmd.ProcessState.ExitCode() instead.
 func CmdExitCodeFromError(err error) (int, error) {
 	cmdExitCode := 0
 	if err != nil {
 		if exitError, ok := err.(*exec.ExitError); ok {
+			exitError.ProcessState.Exited()
```
lpusok, line 36: Is this line needed?

lpusok (changes requested):

errorutil/errorutil.go
godrei: oh no, definitely not

command/command_test.go
```diff
@@ -1,49 +1,108 @@
-package command
+package command_test
```
mateherber, line 1: Is this a best practice to put tests to a different package than testee?

mateherber (dismissed):
LGTM

command/command_test.go
godrei: this way you do blackbox testing and enforce testing the exported functions only

lpusok (approved):

mateherber (approved):

PR #113, opened 2021-03-16
Support extra arguments for SubmoduleUpdate

adborbas:
### Description
Add `opts` argument to `SubmoduleUpdate` function on gitcommand to support extra args from caller.

### DIFF
--- a/command/git/commands.go
+++ b/command/git/commands.go
@@ -62,9 +62,11 @@ func (g *Git) Clean(options ...string) *command.Model {
 }
 
 // SubmoduleUpdate updates the registered submodules.
-func (g *Git) SubmoduleUpdate(opts ...string) *command.Model {
+func (g *Git) SubmoduleUpdate(shallowCheckout bool) *command.Model {
 	args := []string{"submodule", "update", "--init", "--recursive"}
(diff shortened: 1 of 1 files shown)


lpusok:
What do you think about only exposing the needed (depth) parameters, instead of arbitrary arguments? That would help separating the concerns of creating the command and setting the configuration.

adborbas:
I've thought of this as well, but went with `opts...` as:

- it makes any non-required arguments optional,
- have the save behaviour as `Fetch`, `Clean` and `Branch` functions.

How would you define the signature of the function to only expose the `depth` paramater?

lpusok:
> I've thought of this as well, but went with `opts...` as:
> 
>     * it makes any non-required arguments optional,
> 
>     * have the save behaviour as `Fetch`, `Clean` and `Branch` functions.
> 
> 
> How would you define the signature of the function to only expose the `depth` paramater?

SubmoduleUpdate(shallowCheckout bool) OR just hardcode the depth=1 in ShallowSubmoduleUpdate().

adborbas:
@lpusok done, please review again.

lpusok (approved):

PR #114, opened 2021-03-18
Use --jobs flag on fetch and submodule update commands

istvankovacs-bitrise:
https://bitrise.atlassian.net/browse/STEP-680

### Description
`--jobs` flag with the default value of `10` added to the `git fetch` and `git submodule update` commands, based on [this](https://bitrise.atlassian.net/wiki/spaces/STEP/pages/1325236987/Git+Clone+Step+--jobs+flag+investigation) investigation.

### DIFF
--- a/command/git/commands.go
+++ b/command/git/commands.go
@@ -62,8 +62,9 @@ func (g *Git) Clean(options ...string) *command.Model {
 }
 
 // SubmoduleUpdate updates the registered submodules.
-func (g *Git) SubmoduleUpdate(shallowCheckout bool) *command.Model {
-	args := []string{"submodule", "update", "--init", "--recursive", "--jobs=10"}
+func (g *Git) SubmoduleUpdate(shallowCheckout bool, opts ...string) *command.Model {
+	args := []string{"submodule", "update", "--init", "--recursive"}
+	args = append(args, opts...)
 	if shallowCheckout {
 		args = append(args, "--depth=1")
 	}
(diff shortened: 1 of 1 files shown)


command/git/commands.go
```diff
@@ -31,7 +31,7 @@ func (g *Git) RemoteAdd(name, url string) *command.Model {
 
 // Fetch downloads objects and refs from another repository.
 func (g *Git) Fetch(opts ...string) *command.Model {
-	args := []string{"fetch"}
+	args := []string{"fetch", "--jobs=10"}
```
mateherber, line 34: Are we sure we want to hardcode 10 here?

mateherber (changes requested):

command/git/commands.go
istvankovacs-bitrise: Based on the [action points](https://bitrise.atlassian.net/wiki/spaces/STEP/pages/1325236987/Git+Clone+Step+--jobs+flag+investigation#Suggested-follow-up) of the investigation, I believe yes.

Since it is only used for two commands, I don't see a point in extracting it to a const.

mateherber: I see but this is not the git clone step but a "generic" utility kind of lib. I think hardcoding 10 in git clone step might make sense however here I'd advise against it. wdyt?

istvankovacs-bitrise: Makes sense! `fetch` already has an additional args, so we could just pass the `jobs` from the `git-clone` Step, and we can create a similar input for the `submodule update` as well.

mateherber: maybe introducing an overload FetchWithJobs() would be the safest choice. And the standard fetch just calls it with jobs = 0 and it adds the jobs argument only if it is more than 0.

That way we don't break the api

mateherber: Cool

mateherber (approved):

PR #115, opened 2021-03-22
Add sparse-checkout commands to Git

adborbas:
### Description

Add sparse-checkout commands to Git.

More info in [git-sparse-checkout](https://git-scm.com/docs/git-sparse-checkout)

### Changes
- Adding `SparseCheckoutInit` and `SparseCheckoutSet` functions to `git.commands`.
- Added tests.

### DIFF
--- a/command/git/commands.go
+++ b/command/git/commands.go
@@ -145,12 +145,18 @@ func (g *Git) Config(key string, value string) *command.Model {
 	return g.command("config", key, value)
 }
 
-func (g *Git) SparseCheckoutInit() *command.Model {
-	return g.command("sparse-checkout", "init", "--cone")
+// SparseCheckoutInit initializes the sparse-checkout config file.
+func (g *Git) SparseCheckoutInit(cone bool) *command.Model {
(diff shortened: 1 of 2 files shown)


command/git/commands.go
```diff
@@ -144,3 +144,19 @@ func (g *Git) Status(opts ...string) *command.Model {
 func (g *Git) Config(key string, value string) *command.Model {
 	return g.command("config", key, value)
 }
+
+// SparseCheckoutInit initializes the sparse-checkout config file.
+func (g *Git) SparseCheckoutInit(cone bool) *command.Model {
```
mateherber, line 149: Why not the usual `opts ...string` Is this `--cone` special?

adborbas: I have seen both ways used here and since `--cone` is the only flag for [init](https://git-scm.com/docs/git-sparse-checkout#Documentation/git-sparse-checkout.txt-eminitem) I've went with this option.

lpusok (approved):

PR #130, opened 2021-07-13
Adding additional options parameter to git log command

lpusok:
### Context

Added option to specify revision for the git log command.

Resolves: https://bitrise.atlassian.net/browse/STEP-1135

### Changes

- Added additional options paramter to git log command.


Bence1001 (approved):
LGTM!

PR #131, opened 2021-08-05
Add basic building blocks of the Ideal Step Architecture

Bence1001:
## Context
* We came up with a new, testable architecture for steps: the [Ideal Step Architecture](https://bitrise.atlassian.net/l/c/91yzFT1u).
* We will be migrating all steps to this testable architecture eventually.
* There are reusable, basic building blocks of this architecture (e.g. for running a command or setting environment variables) that should be added to a common library: `go-utils` and `go-steputils`.

## Changes
**Added basic building blocks of the Ideal Step Architecture.**
* Based on this draft PR that migrates the Activate SSH key step to the Ideal Step Architecture: https://github.com/bitrise-steplib/steps-activate-ssh-key/pull/25.
* **Added** `Command`: represents a runnable CLI command. Essentially wraps `exec.Cmd`.
* **Added** `OsRepository`:a repository for getting and setting environment variables in the OS. Essentially wraps `export`.
* **Added** `FileWriter`:can write into files.
* **Added** `Logger`: can log to the command line.
* **Added** `TmpDirProvider`: can create a temporary directory.

**Some of the existing functionality have been made deprecated or obsolete in favor of these new interfaces. Details are in the sections below.**

### Command
* `command.Model` has been made internal and renamed to `cmdWrapper` in order to promote the usage of `Command`. `cmdWrapper` is the internal implementation of `Command`.
* Replaced "public" usages of `command.Model` with `Command`.
* Deprecated functions that were only used in 1 or 2 external repositories (e.g. https://github.com/bitrise-io/bitrise) in order to push external repositories to use `Command`.
* Made functions internal that had no usages in external repositories in order to minimize the public API and push external repositories to use `Command`.
* Removed functions that had no usages _at all_ in order to minimize the public API and push external repositories to use `Command`.

### Logger
* Removed existing `log.Logger` interfaces and _all_ of its implementations as it only had 2 usages (both in https://github.com/bitrise-io/bitrise) in order to avoid name clash with the new `log.Logger` and provide a broader functionality.
* Removed `Successf` and `success` severity as it was exactly the same as `Donef` and `done`, but had far fewer usages in external repositories.


command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
 	return strings.Join(cmdArgsDecorated, " ")
 }
 
-// RunCmdAndReturnExitCode ...
-func RunCmdAndReturnExitCode(cmd *exec.Cmd) (exitCode int, err error) {
```
Bence1001, line 139: Made internal because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCmdAndReturnExitCode%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
+// Deprecated: Use Command instead.
+func runCmdAndReturnExitCode(cmd *exec.Cmd) (exitCode int, err error) {
 	err = cmd.Run()
 	exitCode = cmd.ProcessState.ExitCode()
 	return
 }
 
-// RunCmdAndReturnTrimmedOutput ...
-func RunCmdAndReturnTrimmedOutput(cmd *exec.Cmd) (string, error) {
```
Bence1001, line 146: Made internal because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCmdAndReturnTrimmedOutput%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
+// Deprecated: Use Command instead.
+func runCmdAndReturnTrimmedOutput(cmd *exec.Cmd) (string, error) {
 	outBytes, err := cmd.Output()
 	outStr := string(outBytes)
 	return strings.TrimSpace(outStr), err
 }
 
-// RunCmdAndReturnTrimmedCombinedOutput ...
-func RunCmdAndReturnTrimmedCombinedOutput(cmd *exec.Cmd) (string, error) {
```
Bence1001, line 153: Made internal because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCmdAndReturnTrimmedCombinedOutput%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
+// Deprecated: Use Command instead.
+func runCmdAndReturnTrimmedCombinedOutput(cmd *exec.Cmd) (string, error) {
 	outBytes, err := cmd.CombinedOutput()
 	outStr := string(outBytes)
 	return strings.TrimSpace(outStr), err
 }
 
-// RunCommandWithReaderAndWriters ...
-func RunCommandWithReaderAndWriters(inReader io.Reader, outWriter, errWriter io.Writer, name string, args ...string) error {
```
Bence1001, line 160: Removed because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCommandWithReaderAndWriters%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
+// Deprecated: Use Command instead.
 func RunCommandWithWriters(outWriter, errWriter io.Writer, name string, args ...string) error {
 	cmd := exec.Command(name, args...)
 	cmd.Stdout = outWriter
 	cmd.Stderr = errWriter
 	return cmd.Run()
 }
 
-// RunCommandInDirWithEnvsAndReturnExitCode ...
-func RunCommandInDirWithEnvsAndReturnExitCode(envs []string, dir, name string, args ...string) (int, error) {
```
Bence1001, line 177: Inlined because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCommandInDirWithEnvsAndReturnExitCode%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
-	}
 	if len(envs) > 0 {
 		cmd.Env = envs
 	}
 
-	return RunCmdAndReturnExitCode(cmd)
-}
-
-// RunCommandInDirAndReturnExitCode ...
-func RunCommandInDirAndReturnExitCode(dir, name string, args ...string) (int, error) {
```
Bence1001, line 193: Remove because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCommandInDirAndReturnExitCode%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
-}
-
-// RunCommandWithEnvsAndReturnExitCode ...
-func RunCommandWithEnvsAndReturnExitCode(envs []string, name string, args ...string) (int, error) {
-	return RunCommandInDirWithEnvsAndReturnExitCode(envs, "", name, args...)
+	return runCmdAndReturnExitCode(cmd)
 }
 
-// RunCommandInDir ...
-func RunCommandInDir(dir, name string, args ...string) error {
```
Bence1001, line 203: Inlined because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCommandInDir%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
+// RunCommandAndReturnStdout ...
+// Deprecated: Use Command instead.
 func RunCommandAndReturnStdout(name string, args ...string) (string, error) {
 	cmd := exec.Command(name, args...)
-	return RunCmdAndReturnTrimmedOutput(cmd)
+	return runCmdAndReturnTrimmedOutput(cmd)
 }
 
-// RunCommandInDirAndReturnCombinedStdoutAndStderr ...
-func RunCommandInDirAndReturnCombinedStdoutAndStderr(dir, name string, args ...string) (string, error) {
```
Bence1001, line 226: Inlined because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunCommandInDirAndReturnCombinedStdoutAndStderr%22&type=Code) in external repositories.

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
-
-// RunCommandAndReturnCombinedStdoutAndStderr ..
+// RunCommandAndReturnCombinedStdoutAndStderr ...
+// Deprecated: Use Command instead.
 func RunCommandAndReturnCombinedStdoutAndStderr(name string, args ...string) (string, error) {
-	return RunCommandInDirAndReturnCombinedStdoutAndStderr("", name, args...)
-}
-
-// RunBashCommand ...
-func RunBashCommand(cmdStr string) error {
```
Bence1001, line 240: Removed because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunBashCommand%22&type=Code).

command/command.go
```diff
@@ -135,118 +157,70 @@ func PrintableCommandArgs(isQuoteFirst bool, fullCommandArgs []string) string {
...
-	return RunCommand("bash", "-c", cmdStr)
-}
-
-// RunBashCommandLines ...
-func RunBashCommandLines(cmdLines []string) error {
```
Bence1001, line 245: Removed because it was [unused](https://github.com/search?q=org%3Abitrise-io+org%3Abitrise-steplib+%22RunBashCommand%22&type=Code).

env/env.go
```diff
@@ -0,0 +1,32 @@
+package env
+
+import "os"
+
+// OsRepository ...
+type OsRepository interface {
```
Bence1001, line 6: I think this name will be too ambiguous, even if used together with package name `env.OsRepository`.

I would much rather put this in the `os` package and name it `EnvironmentVariableRepository`: `os.EnvironmentVariableRepository`.
This way, the envman-based repository would read like this: `envman.EnvironmentVariableRepository`.

Wdyt?

command/command.go
```diff
@@ -9,121 +9,143 @@ import (
...
+	RunAndReturnTrimmedOutput() (string, error)
+	RunAndReturnExitCode() (int, error)
+	RunAndReturnTrimmedCombinedOutput() (string, error)
+}
+
+// Factory ...
+type Factory func(name string, args ...string) Command
+
+// NewCommand ...
+func NewCommand(name string, args ...string) Command {
```
mateherber, line 31: I think the name of this constructor should be NewXYZCommand like NewRealCommand NewDefaultCommand NewOsCommand, NewWhateverCommand

command/command.go
```diff
@@ -9,121 +9,143 @@ import (
 	"strings"
 )
 
-// ----------
-
-// Model ...
-type Model struct {
+// Command ...
+type Command interface {
+	SetDir(dir string) Command
```
mateherber, line 14: Why do we keep this builder type configuration that seems kinda antipattern in Go instead of adding these to the constructor and factory method?

command/command.go
mateherber: Also seeing other constructors below maybe Command is not needed at the end so from the outside you can write:
`command.NewDefault()` and for the rest `command.NewDefaultWithParams()`.

mateherber: We will never probably use the constructor functions directly just in place of the Factory if we want to keep the NewWithXYZ constructors we probably also need to introduce FactoryWithXYZ for them

command/command.go
mateherber: Hmm thinking more maybe it is worth keeping the builder for these "optional" type of arguments so we avoid polluting the factory methods / constructors. (No default and named values in golang luckily...)

env/env.go
mateherber: I agree. Also I think the term envar (not envvar) is in the lingo enough that we can use it instead of the bloated one: 
`os.EnvarRepository` `envman.EnvarRepository`, wdyt?

fileutil/fileutil.go
```diff
@@ -4,13 +4,46 @@ import (
...
 
+// FileWriter ...
+type FileWriter interface {
+	Write(path string, value string, mode os.FileMode) error
+}
+
+type osFileWriter struct{}
+
+// NewOsFileWriter ...
+func NewOsFileWriter() FileWriter {
```
mateherber, line 22: The "os" term just been used as a shitty replacement for "default" or "real" we should come up with a consistent naming for the "real" implementations that doesn't smell like shit :) I mean I'm kinda happier with `defaultFileWriter/realFileWriter` than `OsFileWriter`.

What I would be even more happier with is `FileWriter` and `IFileWriter` for that matter :)

log/log.go
```diff
@@ -7,6 +7,60 @@ import (
...
+	Debugf(format string, v ...interface{})
+	Errorf(format string, v ...interface{})
+	Println()
+}
+
+type defaultLogger struct {
+}
+
+// NewDefaultLogger ...
+func NewDefaultLogger() Logger {
```
mateherber, line 25: Here we used the "default", see my comments above

mateherber (changes requested):

env/env.go
Bence1001: I'm more in favor of unabbreviated, readable names (e.g. `EnvironmentVariableRepository`) because they are clear to everybody in any context (e.g. new joiners or members of other teams). But if others think it's in the Bitrise lingo then I'm okay with that too. 👍

command/command.go
Bence1001: I think in this case `NewCommand` is fine because there's no missing context here that would justify any extra terms (e.g. `Default` or `OS`). `Command` should be clear to anybody in any context, so I would say there's no need for the extra verbosity.

I forgot to raise it, but I also wanted to ask whether I should rename the existing constructors to contain the word `Command`?

command/command.go
Bence1001: Yeah, I think until we don't have default parameters AND method overloading, adding a new constructor for all combinations would be very painful.

log/log.go
Bence1001: Should we replace it with something more meaningful that adds more context, e.g. `NewCommandLineLogger`?

I can imagine a `Logger` can have multiple implementations in the future (e.g. something that logs to an arbitrary data collector/processor service) so in that sense `Default` would be too ambiguous.

fileutil/fileutil.go
Bence1001: If there's no other possible implementation of `FileWriter`, we can simply call this constructor `NewFileWriter` (see my other comment above about naming these kinds of interfaces/functions). And for the implementing `struct`, we can simply call it `fileWriter`.

env/env.go
mateherber: I assumed it is more than Bitrise lingo, like linux / generic OS lingo. `EnviromentVariable` seems painfully long if everybody knows what `Envar` is. I'm fine either way tho.

fileutil/fileutil.go
mateherber: Not sure having fileWriter and FileWriter is wise. Other than that I agree.

log/log.go
mateherber: Agree with a more specific name is preferable.

env/env.go
Bence1001: After googling a bit, I'm slightly more convinced that `Envar` is a common word, but I'm still not entirely sure. 😄  Let's see what others think!

ofalvai: I'm in favor of abbreviations where it's trivial to understand from the context, and I think this is the case here (Bitrise steps are all about env vars). But I don't like `Envar`, it seems like a typo to me 😅

command/command.go
lpusok: A note that it will be used as command.NewCommand ,so adding Command here can be considered 'stuttering' which is  not recommended. See https://stackoverflow.com/a/45886055

command/command.go
lpusok: Can you please check bitrise-io and bitrise-steplib orgs too? Not sure if GitHub search indexes not recently touched repos so the safest way is to check out all step repos and search in the local file system.
If you want to be extra sure the few ~10 3rd party golang steps may be checked for usages.

Bence1001: I've seen search results in repositories that were archived years ago. For example, [this query](https://github.com/search?q=org%3Abitrise-io+RunCommandAndReturnStdout&type=code) returns `bitrise-bridge`, which was archived in 2019.

Do you think that counts?

command/command.go
Bence1001: @lpusok How about we move it to either `cli` or `os`?

env/env.go
godrei: I also do not like `Envar` but suggest moving the os and envman env vars into different packages.

command/command.go
Bence1001: Discussed offline, we are confident the changes are correct.

env/env.go
Bence1001: Discussed offline, we settled with the following naming scheme:
* `os.EnvironmentRepository` for the OS-based environment variable storage.
* `envman.EnvironmentRepository` for the Envman-based environment variable storage.
* `Get/Set/ListVariable/s()` for function names.
* `defaultEnvironmentRepository` for the implementing `struct` (we are going to apply the same naming convention where interfaces and structs match 1:1).

command/command.go
Bence1001: Discussed offline, we settled with the following naming scheme:
* `command.Command` for the interface.
* `command.New` for the constructor.
* No overloads for the constructor/factory.

command/command.go
Bence1001: Discussed above ☝️

fileutil/fileutil.go
Bence1001: As discussed above, we settled with `defaultFileWriter`.

os/env.go
```diff
@@ -0,0 +1,32 @@
+package os
+
+import "os"
+
+// EnvironmentRepository ...
+type EnvironmentRepository interface {
+	SetVariable(key string, value string) error
```
ofalvai, line 7:
```suggestion
	SetVariable(key, value string) error
```

os/env.go
```diff
@@ -0,0 +1,32 @@
...
+
+type defaultEnvironmentRepository struct{}
+
+// NewEnvironmentRepository ...
+func NewEnvironmentRepository() EnvironmentRepository {
+	return defaultEnvironmentRepository{}
+}
+
+// SetVariable ...
+func (m defaultEnvironmentRepository) SetVariable(key string, value string) error {
```
ofalvai, line 20: Is the receiver intentionally called `m`?

os/env.go
mateherber: I'm not a fan of this suggested syntax.

os/env.go
```diff
@@ -0,0 +1,32 @@
...
+	return os.Setenv(key, value)
+}
+
+// UnsetVariable ...
+func (m defaultEnvironmentRepository) UnsetVariable(key string) error {
+	return os.Unsetenv(key)
+}
+
+// ListVariable ...
+func (m defaultEnvironmentRepository) ListVariable() []string {
```
ofalvai, line 30: Shouldn't this be called `ListVariables()`?

command/command.go
```diff
@@ -9,121 +9,154 @@ import (
...
+		return newCmdWrapper(params[0]), nil
 	}
 
-	return New(params[0], params[1:]...), nil
+	return newCmdWrapper(params[0], params[1:]...), nil
 }
 
 // NewFromSlice ...
-func NewFromSlice(slice []string) (*Model, error) {
+// Deprecated.
```
mateherber, line 66: I'd just delete the "deprecated" stuff. We'll update the usage sites when necessary. Other than that it is vendored

os/env.go
Bence1001: +1, I still get confused when I read something like this in function parameter lists.

os/env.go
Bence1001: Good catch, I'll rename it. Do you have a suggestion?

os/env.go
Bence1001: Good catch, I'll rename it!

command/command.go
Bence1001: I'm okay with both approaches, but I can imagine with this many deprecations it would be overwhelming for some external repositories to migrate away from all methods/functions that would be deleted by this.

We can delete them, but I'd resort to deprecations for now. Those are correctly highlighted by IDEs as well, so they hardly go under the radar.

os/env.go
lpusok: _r_ 😁

os/env.go
mateherber: https://github.com/mvdan/gofumpt#extra-rules-behind--extra
There is an "extended" formatter which has this as en extra option. 

That being said I still think it makes the code less readable so I'd be happy to write type for all params

ofalvai: Let's agree to use the full syntax than, I'm okay with it.

ofalvai (dismissed):

command/command.go
```diff
@@ -1,129 +1,113 @@
...
-func NewFromSlice(slice []string) (*Model, error) {
-	return NewWithParams(slice...)
-}
-
-// NewWithCmd ...
-func NewWithCmd(cmd *exec.Cmd) *Model {
-	return &Model{
-		cmd: cmd,
+// Create ...
+func (d defaultFactory) Create(name string, args []string, opts *Opts) Command {
```
Bence1001, line 34: Nit: I'd call this receiver `f` because `d` is confusing at the call site (I always have to look it up).

command/command.go
```diff
@@ -1,129 +1,113 @@
...
+type defaultCommand struct {
+	cmd *exec.Cmd
 }
 
-// SetStdout ...
-func (m *Model) SetStdout(out io.Writer) *Model {
-	m.cmd.Stdout = out
-	return m
+// NewCommand ...
+func NewCommand(cmd *exec.Cmd) Command {
```
Bence1001, line 64: What's the use case for this constructor?

os/env.go
```diff
@@ -0,0 +1,32 @@
+package os
+
+import "os"
+
+// EnvironmentRepository ...
+type EnvironmentRepository interface {
```
Bence1001, line 6: @mateherber Don't forget to remove this!

Bence1001 (commented):
@mateherber left some comments, mostly minor stuff, but there's one thing left before we merge this PR.

lpusok (approved):

command/command.go
mateherber: It is used in the test right now, but I agree that maybe in real world we don't want to wrap standard commands to our

lpusok (dismissed):

command/command.go
lpusok: It may be useful to wrap a go command in the new one to avoid a big refactor, in the bitrise CLI repo for example I remember doing something  that.

mateherber: Just do the refactor :) We broke the entire command package anyway

lpusok: :D I must agree you are right.

mateherber (approved):

PR #142, opened 2021-09-03
Introduce FileRemover

godrei:
This PR introduces FileRemover to break the direct dependency on `os.Remove` and `os.RemoveAll`.


mateherber (approved):

fileutil/fileutil.go
```diff
@@ -4,13 +4,37 @@ import (
...
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+
+	"github.com/bitrise-io/go-utils/pathutil"
 )
 
+// FileRemover ...
+type FileRemover interface {
```
Bence1001, line 16: Do we have other interfaces in `go-utils` that operate on files? Should we rename this or merge it with existing interfaces? I expect other file operations in the future too.

fileutil/fileutil.go
```diff
@@ -4,13 +4,37 @@ import (
...
+	Remove(name string) error
+	RemoveAll(path string) error
+}
+
+type defaultFileRemover struct{}
```
Bence1001, line 21: You can remove `default` now. 😏

Bence1001 (changes requested):
I have two smaller comments.

PR #153, opened 2022-01-25
Implement analytics logger

mateherber:
# Context

As a part of the data overhaul project we need an easy-to-use API in our steps and tools that sends data to our [analytics service](https://github.com/bitrise-io/bitrise-step-analytics).

# Changes

- Created `Tracker` interface which can be used to send analytics events to our [analytics service](https://github.com/bitrise-io/bitrise-step-analytics)
  - Use `Wait()` method to wait for all job processed
- Created `Client` interface which is responsible for network communication


analytics/client.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"net/http"
+	"time"
+
+	"github.com/bitrise-io/go-utils/retry"
+)
+
+const trackEndpoint = "https://bitrise-step-analytics.herokuapp.com/track"
+const timeOutInSecs = 30
```
godrei, line 12: this would be simpler: `const timeout = 30 * time.Second`

analytics/client_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func Test_trackerClient_send(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		b, _ := io.ReadAll(req.Body)
+		assert.Equal(t, string(b), "{}")
+		assert.Equal(t, req.Method, http.MethodPost)
```
godrei, line 16: NOTE: might be worth adding a test for the content type of the request

analytics/client_test.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"bytes"
+	"github.com/stretchr/testify/assert"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func Test_trackerClient_send(t *testing.T) {
```
godrei, line 12: NOTE: might be worth adding a failing test case too.

analytics/client_test.go
```diff
@@ -0,0 +1,23 @@
+package analytics
+
+import (
+	"bytes"
```
godrei, line 4: forgot to run `goimports`?

analytics/track.go
```diff
@@ -0,0 +1,80 @@
+package analytics
+
+import (
+	"bytes"
+	"encoding/json"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
```
ofalvai, line 9: Shouldn't we use the other UUID package we agreed on in the RFC? https://bitrise.atlassian.net/wiki/spaces/ENGI/pages/2122547554/Standardize+used+uuid+package+for+Go+projects#%F0%9F%A4%94-Why?

analytics/track.go
```diff
@@ -0,0 +1,80 @@
...
+	sharedProperties map[string]interface{}
+}
+
+// NewDefaultTracker ...
+func NewDefaultTracker(shared map[string]interface{}) Tracker {
+	return NewTracker(make(chan *bytes.Buffer, bufferSize), &sync.WaitGroup{}, poolSize, NewDefaultClient(), shared)
+}
+
+// NewTracker ...
+func NewTracker(jobs chan *bytes.Buffer, waitGroup *sync.WaitGroup, size int, client Client, sharedProperties map[string]interface{}) Tracker {
```
godrei, line 34: do we need to wire in anything else than `client` and `sharedProperties` ? think neither in production nor in the test will we alter these.

godrei: NOTE: suggest unifying the var name of `sharedProperties` (I suggest `shared`) to clarify: both sets the same field of the struct.

analytics/track.go
```diff
@@ -0,0 +1,80 @@
...
+	t := tracker{jobs: jobs, waitGroup: waitGroup, client: client, sharedProperties: sharedProperties}
+	t.init(size)
+	return &t
+}
+
+// Enqueue ...
+func (t tracker) Enqueue(eventName string, properties map[string]interface{}) {
```
godrei, line 41: why not pass an `Event` here? by doing so later we could easier alter the Event struct (default formats and fields can be handled with a constructor for the Event struct)

godrei: might Event should be an interface, which can be JSONMarshalled, and shared fields can be set on it.

analytics/track.go
ofalvai, line 34: I feel like it's not a good idea to let the caller decide the "pool size". Is this really expected that steps provide different size values? Why don't we just define a reasonable default value here, and later add a new constructor with a size parameter if needed?

mateherber: there is a default constructor that pins down that as well.

analytics/client_test.go
mateherber: what is that?

mateherber: Oh you mean the order of the imports?

analytics/track.go
mateherber: Didn't know about this 😳 will be using this then 👍🏻

analytics/track.go
mateherber: Ok I'll hardcode the pool mechanism

analytics/track.go
mateherber: I wouldn't set the shared properties on the Event, other than that it sounds ok

ofalvai (dismissed):

godrei (dismissed):

analytics/event.go
```diff
@@ -8,46 +8,28 @@ import (
...
-func (e eventBuilder) AddProperty(key string, value interface{}) EventBuilder {
-	e.properties[key] = value
-	return e
-}
-
-// Build ...
-func (e eventBuilder) Build() Event {
-	return newEvent(e.eventName, e.properties)
+// EventDTO ...
+type EventDTO struct {
```
ofalvai, line 12: DTO as in Data Transfer Object? Is this Go lingo for data classes?

analytics/property.go
```diff
@@ -0,0 +1,131 @@
...
+	GetValue() interface{}
+}
+
+type stringProperty struct {
+	key   string
+	value string
+}
+
+// StringProperty ...
+func StringProperty(key string, value string) Property {
```
ofalvai, line 15: So much of this file will be rewritten once Go 1.18 ships with generics 😬

ofalvai (dismissed):

analytics/event.go
mateherber: It's Java lingo for serializable classes 😅

analytics/property.go
mateherber: Every Go file ever written will be rewritten after generics will be a thing 😛

godrei (approved):

godrei (dismissed):

analytics/track.go
```diff
@@ -2,62 +2,57 @@ package analytics
...
-		go t.worker()
-	}
-}
-
-func (t tracker) worker() {
-	for job := range t.jobs {
-		t.client.Send(job)
-		t.waitGroup.Done()
-	}
+// Fork ...
```
ofalvai, line 55:
```suggestion
// Pin ...
```

ofalvai (dismissed):
LGTM, approved (but for the last time 😆 It's good enough.)

ofalvai (dismissed):

PR #158, opened 2022-04-14
Helper method to transform a json annotated struct to Properties

mateherber:


analytics/property.go
```diff
@@ -0,0 +1,43 @@
...
+
+// AppendIfNotEmpty ...
+func (p Properties) AppendIfNotEmpty(key string, value string) {
+	if value != "" {
+		p[key] = value
+	}
+}
+
+// ToProperty Converts struct into Properties map based on json tags
+func ToProperty(item interface{}) Properties {
```
tothszabi, line 30: nitpick: I would propose to rename this function to `newProperty` because from the call site it looks like a constructor and that is how we called them for the ideal step refactor.

tothszabi (approved):
Looks great

analytics/property_test.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		{
+			name: "Empty Test",
+			item: emptyTestStruct{},
+			want: Properties{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equalf(t, tt.want, ToProperty(tt.item), "ToProperty(%v)", tt.item)
```
tothszabi, line 78: There are two leftover `ToProperty` names here which can be renamed.

tothszabi (approved):

PR #159, opened 2022-07-14
Add runtime checks for CPU architecture

hisaac:
## Context

I'm working on updating the avd-manager and wait-for-android-emulator steps to have some conditional behavior based on the CPU-type they are being run on, so I needed a way to check the CPU-type at runtime. This seemed like something that could be useful outside of these two Steps, so I figured I'd add it to go-utils for other consumers to use as well.

Here is an example of how this change could be used within a Step:

```go
cpuIsARM, err := system.CPU.IsARM()
if err != nil {
	log.Errorf("Failed to check CPU: %s", err)
} else if cpuIsARM {
	log.Warnf("This Step is not yet supported on Apple Silicon (M1) machines. If you're unable to find a solution to this error, try running this Workflow on an Intel-based machine type.")
}
```

From here: https://github.com/bitrise-steplib/steps-avd-manager/blob/c175c0975ac38fdaef32051e39e8e1d240f7df64/main.go#L86-L91

## Changes

- Creates a `system` module with a `cpu` type.
- Adds methods to check the CPU architecture, and a convenience method to return a boolean of if the CPU is arm or not.
- Adds a couple tests of the new module.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,4 +7,5 @@ require (
 	github.com/gofrs/uuid v4.2.0+incompatible
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/sys v0.0.0-20220712014510-0a85c31ab51e
 )
--- a/system/cpu.go
+++ b/system/cpu.go
@@ -8,6 +8,7 @@ import (
 
 type cpu struct{}
 
+// CPU is a singleton that returns information about the CPU
 var CPU cpu
 
 // Architecture returns the CPU's architecture name as a string
@@ -31,7 +32,7 @@ func (cpu) IsARM() (bool, error) {
 
 	if err != nil {
 		return false, err
-	} else {
-		return strings.Contains(cpuType, "arm"), nil
 	}
+
+	return strings.Contains(cpuType, "arm"), nil
 }
--- a/system/cpu.go
+++ b/system/cpu.go
@@ -12,6 +12,10 @@ type cpu struct{}
 var CPU cpu
 
 // Architecture returns the CPU's architecture name as a string
+//
+// Some example return values:
+//   - "x86_64"
+//   - "arm64"
 func (cpu) Architecture() (string, error) {
 	var utsname unix.Utsname
 	err := unix.Uname(&utsname)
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -7,4 +7,5 @@ require (
 	github.com/gofrs/uuid v4.2.0+incompatible
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/sys v0.0.0-20220712014510-0a85c31ab51e
```
russellbstephens, line 10: is there something less specific to rely on?

hisaac: Not sure what you mean by less-specific. If you mean, could we target a tag instead, these system packages don't appear to have tagged versions. If you look on the right side of the page here, there isn't anything in the Tagged Versions list. https://pkg.go.dev/golang.org/x/sys

system/cpu.go
```diff
@@ -0,0 +1,38 @@
...
+// Architecture returns the CPU's architecture name as a string
+func (cpu) Architecture() (string, error) {
+	var utsname unix.Utsname
+	err := unix.Uname(&utsname)
+	if err != nil {
+		return "", err
+	}
+
+	architecture := string(utsname.Machine[:])
+	architecture = strings.Trim(architecture, "\x00")
```
hisaac, line 23: The value returned from `utsname.Machine[:]` is a string with a bunch of `NUL` values after it for some reason. So this strips those off the end.

system/cpu.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+type cpu struct{}
+
+// CPU is a singleton that returns information about the CPU
+var CPU cpu
```
russellbstephens, line 12: i like this aliasing

russellbstephens (dismissed):

system/cpu.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Architecture returns the CPU's architecture name as a string
```
ofalvai, line 14: Can you add an example return value to this description, please? Looking at the code for the first time and seeing that the return type is a plain string, I have a hard time imagining what will I get by calling this function.

ofalvai (dismissed):
💯

system/cpu.go
hisaac: Great idea, will do.

PR #162, opened 2022-08-02
Add retryable HTTP client from v1

ofalvai:
Moves our preconfigured retryable HTTP client [in v1](https://github.com/bitrise-io/go-utils/blob/v1/retry/retryhttp.go) to the current master.

### DIFF
--- a/http/retryhttp.go
+++ b/http/retryhttp.go
@@ -5,21 +5,21 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
-// HTTPLogAdaptor adapts the retryablehttp.Logger interface to the go-utils logger.
-type HTTPLogAdaptor struct {
+// NewHTTPClient returns a retryable HTTP client with common defaults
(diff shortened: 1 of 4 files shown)


http/retryhttp.go
```diff
@@ -0,0 +1,25 @@
+package http
+
+import (
+	"github.com/bitrise-io/go-utils/v2/log"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// NewHTTPClient returns a retryable HTTP client with common defaults
+func NewHTTPClient(logger log.Logger) *retryablehttp.Client {
```
godrei, line 9: Clients will use this function this way: `http.NewHTTPClient`, while earlier this was: `retry.NewHTTPClient`. I prefer the older syntax, as it gives more context to this package (**retry**).

What about using `httpretry` as package name (-> `httpretry.NewClient`) or using the original package name (`retry` -> `retry.NewHTTPClient`)?

godrei (approved):

PR #165, opened 2022-08-11
Remove go-utils/v1 dependency

ofalvai:
This removes the last remaining circular dependencies from this library.

Changes:
- Make `analytics/client` use the retry HTTP client from `retryhttp`
- Move the `colorstring` package over from v1 as `log/colorstring`. Everying that was public in v1 is still public in v2.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,19 @@
 module github.com/bitrise-io/go-utils/v2
 
-go 1.16
+go 1.17
 
 require (
-	github.com/bitrise-io/go-utils v1.0.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,12 +1,19 @@
 module github.com/bitrise-io/go-utils/v2
 
-go 1.16
+go 1.17
 
 require (
-	github.com/bitrise-io/go-utils v1.0.1
```
ofalvai, line 6: 🎉 🎉 🎉

lpusok (approved):

PR #166, opened 2022-09-14
[PLANG-2394] Add advanced logger

tothszabi:
# Changes

This PR adds a new advanced logger which can either deliver the logs as structured json messages or the existing human readable format.

# Implementation details

There are three new loggers:
- Main
- Json
- Legacy

The main logger is the entry point for this new system. This is the highest level logger and acts as wrapper so the internal actual logger implementation can be replaced during the runtime. So the main logger just forwards the function calls to the internal one. This main logger can be invoked via global logging functions (to have some backwards compatibility) or the public global instance.

One actual implementation is the json logger which just simply converts the parameters into the predefined messages. Currently just a single log message is implemented but the other types will follow soon.

The other concrete implementation is the legacy logger which just simply wraps the already existing log functionality we have in this package.

I want to also mention the use case for the `advancedlog/writer_test.go` file. It is not obvious here because it will be used only by the cli. This writer helps capturing the stdout and stderr outputs when executing a step or a cli plugin. In the cli the function callback will just simply forward the captured output to the new advanced logger.


advancedlog/logger.go
```diff
@@ -0,0 +1,60 @@
+package logger
+
+import "os"
+
+const (
+	// OutputFormatKey ...
+	OutputFormatKey = "output-format"
```
godrei, line 7: Do we need this here?

advancedlog/types.go
```diff
@@ -0,0 +1,55 @@
+package logger
+
+// Producer ...
+type Producer int
```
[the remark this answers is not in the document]
tothszabi: This is a great idea and I have admit I did not think of it at all.

advancedlog/types.go
```diff
@@ -0,0 +1,55 @@
...
+	case WarnLevel:
+		return "warn"
+	case DoneLevel:
+		return "done"
+	case NormalLevel:
+		return "normal"
+	case DebugLevel:
+		return "debug"
+	default:
+		return "info"
```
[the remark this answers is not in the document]
tothszabi: `normal` does sound better as the default.I will change this.

advancedlog/logger.go
tothszabi: No, these are the parameters for the cli. I forgot to remove them when I moved the logger into this package.

advancedlog/json_logger.go
```diff
@@ -0,0 +1,57 @@
...
+		Timestamp:   j.timeProvider().Format(RFC3339Micro),
+		MessageType: "log",
+		Producer:    string(producer),
+		Level:       string(level),
+		Message:     message,
+	}
+
+	err := j.encoder.Encode(logMessage)
+	if err != nil {
+		// This is only to satisfy errcheck.
```
godrei, line 55: Shouldn't we introduce better error handling for this, like printing the error to the build log using a hand-crafted json?

advancedlog/legacy_logger.go
```diff
@@ -0,0 +1,53 @@
...
+	return l.debugLogEnabled
+}
+
+func (l *legacyLogger) LogMessage(producer Producer, level Level, message string) {
+	if l.debugLogEnabled == false && level == DebugLevel {
+		return
+	}
+
+	// It is needed to trim the newline char from the end because the wrapped logger will automatically add one
+	if message != "\n" {
```
godrei, line 35: IMO we should use log functions that add newline to the end of the message and we shouldn't modify the original message. With these 2 I think:
- we should remove the newlines on the client side, where it is not needed (where we use old log functions and used a logging method that doesn't add newline at the end, but we added one)
- we should remove this workaround from here

advancedlog/logger.go
```diff
@@ -0,0 +1,52 @@
...
+	EnableDebugLog(enabled bool)
+	IsDebugLogEnabled() bool
+	LogMessage(producer Producer, level Level, message string)
+}
+
+// DefaultLogger ...
+var DefaultLogger = NewMainLogger()
+
+// SetupLogger ...
+func SetupLogger(outputFormat string) {
```
godrei, line 43: IMO we could use a dedicated type for the output format:

```
type OutputFormat string

const (
	// JSONFormat ...
	JSONFormat OutputFormat = "json"
)

...

func SetupLogger(outputFormat OutputFormat) {
...
```

godrei: It seems current you can configure two things on the logger:
- outputFormat
- whether debug mode is enabled

I think they should have similar function names, like:
- SetEnableDebugLog
- SetOutputFormat

advancedlog/main_logger.go
```diff
@@ -0,0 +1,122 @@
...
+	m.internalLogger.LogMessage(CLI, NormalLevel, fmt.Sprintf(format, args...))
+}
+
+// Println ...
+func (m *MainLogger) Println(args ...interface{}) {
+	m.internalLogger.LogMessage(CLI, NormalLevel, fmt.Sprintln(args...))
+}
+
+// TPrintf ...
+func (m *MainLogger) TPrintf(format string, args ...interface{}) {
```
godrei, line 99: Do we really need log functions with `T` prefix? Earlier it added time prefix to messages.

advancedlog/types.go
```diff
@@ -0,0 +1,29 @@
+package logger
+
+// Producer ...
+type Producer string
+
+const (
+	// CLI ...
+	CLI Producer = "cli"
```
godrei, line 8: IMO we shouldn't use `cli` as it is too generic, let's use `bitrise cli`

advancedlog/json_logger.go
tothszabi: That is a good point. I was thinking that none of the parameters are complex in the sense that they are built in types  and we control the message struct so this should never fail in theory. 

But I like the idea to construct a json string manually and print it out.

advancedlog/legacy_logger.go
tothszabi: Okay, you convinced me. Initially I thought that this is not a big deal but if we really want to preserve the original message (like when the developer put a newline at the end on purpose) then we need to remove this.

advancedlog/types.go
tothszabi: Yes, definitely but I will add an underscore (`bitrise_cli`) to it because all of our multi word values have an underscore in them.

advancedlog/main_logger.go
tothszabi: I actually forgot to implement the time prefix which I will add because in the legacy log it will be still useful. For the new json logging I am hoping that we could add a time prefix to every line based on the time value in the json field. That would make the logs way more useful.

tothszabi: Actually there is only one usage in the cli to log the timestamp. I will remove this function.

advancedlog/color_converter.go
```diff
@@ -0,0 +1,42 @@
...
+func convertColoredString(message string) (Level, string) {
+	logLevel := NormalLevel
+
+	for code, level := range ansiEscapeCodeToLevel {
+		if strings.HasPrefix(message, code) {
+			logLevel = level
+			message = strings.TrimPrefix(message, code)
+			hasNewline := strings.HasSuffix(message, "\n")
+			// We need to remove all the possible noise from the end as we need remove the reset ansi code from the end
+			message = strings.TrimRightFunc(message, unicode.IsSpace)
```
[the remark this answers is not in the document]
tothszabi: I like what you are proposing. And I would make it a bit more strict. You asked why are not we trimming the leading space characters. I noticed that our old log coloring wraps the supplied message with the color codes. So if this is the original message (notice the 4 spaces at the beginning)
```
    Original message
```
then the old colorized messages will be 
```
[color-code]    Original message[reset-color-code]
```
So if we encounter a log message with 4 spaces being in front of the color code like
```
    [color-code]Original message
```
then that means that was not produced by our stuff but rather a 3rd party tool / command. So I think if we want to extract the log level from the message then we need to use the following rules:
- starts with a color code
- can have trailing whitespace characters but the first non-whitespace characters have to be part of the reset-color code
- contains exactly one color and reset code

advancedlog/color_converter.go
```diff
@@ -0,0 +1,68 @@
...
+	"\u001b[34;1m": InfoLevel,
+	"\u001b[32;1m": DoneLevel,
+	"\u001b[35;1m": DebugLevel,
+}
+
+func convertColoredString(message string) (Level, string) {
+	logLevel := NormalLevel
+
+	// We need to remove all the possible noise from the end as we need remove the reset ansi code from the end
+	message = strings.TrimRightFunc(message, unicode.IsSpace)
```
godrei, line 24: Is it OK to modify all the logs (even if they are not colored)?

advancedlog/color_converter.go
```diff
@@ -0,0 +1,68 @@
...
+
+var ansiEscapeCodeToLevel = map[string]Level{
+	"\u001b[31;1m": ErrorLevel,
+	"\u001b[33;1m": WarnLevel,
+	"\u001b[34;1m": InfoLevel,
+	"\u001b[32;1m": DoneLevel,
+	"\u001b[35;1m": DebugLevel,
+}
+
+func convertColoredString(message string) (Level, string) {
```
godrei, line 20: Would be good to comment what we consider a colored message ([from this comment](https://github.com/bitrise-io/go-utils/pull/166#discussion_r977686832)):
- starts with a color code
- can have trailing whitespace characters but the first non-whitespace characters have to be part of the reset-color code
- contains exactly one color and reset code pair

advancedlog/legacy_logger.go
```diff
@@ -0,0 +1,56 @@
...
+
+// LogMessage ...
+func (l *legacyLogger) LogMessage(producer Producer, level Level, message string) {
+	if !l.debugLogEnabled && level == DebugLevel {
+		return
+	}
+
+	// All the print functions below automatically add a newline to the end. We need to replace the newline with an
+	// empty line not to log multiple newlines.
+	if message == "\n" {
```
godrei, line 38: I'm still not sure if we should do any transformation on the given log message (if we misused log calls in the CLI [adding an extra newline at the end], we should simply remove them).

advancedlog/logger.go
```diff
@@ -0,0 +1,54 @@
...
+	EnableDebugLog(enabled bool)
+	IsDebugLogEnabled() bool
+	LogMessage(producer Producer, level Level, message string)
+}
+
+// DefaultLogger ...
+var DefaultLogger = NewMainLogger()
+
+// SetOutputFormat ...
+func SetOutputFormat(outputFormat string) {
```
godrei, line 45: `outputFormat` type could be `outputFormat` to ensure clients are using valid values.

advancedlog/color_converter.go
godrei: updated in https://github.com/bitrise-io/go-utils/pull/166/commits/3d6a1120d6943e0fc7f55f6d67bc8eafa6f14e6e

advancedlog/legacy_logger.go
godrei: updated in https://github.com/bitrise-io/go-utils/pull/166/commits/4323136f49bc1342725c43a9d59be3ae0ac2b0f0

advancedlog/logger.go
godrei: Updated in https://github.com/bitrise-io/go-utils/pull/166/commits/38b681bef8d4ec5b938ae7156c6515d6f3bb6cbb

gaborszakacs:
This was flagged by Haystack as a long-running PR, is this still relevant?

tothszabi:
@gaborszakacs No it is not. I will close it.

PR #168, opened 2022-09-16
Use standard linters

ofalvai:
Use the standard linters and config we use in other step repos.

This PR drops errcheck and golint in favor of golangci-lint through the check internal step.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,26 @@
+run:
+  skip-dirs:
+  - .*/mocks
+
+issues:
+  # https://github.com/golangci/golangci-lint/issues/2439
+  exclude-use-default: false
+
+linters:
+  enable:
+  - errcheck
+  - gosimple
+  - govet
+  - ineffassign
+  - staticcheck
+  - typecheck
+  - unused
+  - revive
+
+linters-settings:
+  revive:
+    severity: error
+    rules:
+    - name: exported
+      arguments:
+      - checkPrivateReceivers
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -0,0 +1,26 @@
...
+  - unused
+  - revive
+
+linters-settings:
+  revive:
+    severity: error
+    rules:
+    - name: exported
+      arguments:
+      - checkPrivateReceivers
```
tothszabi, line 26: Are these repo specific settings needed because the global one was producing issues?

tothszabi: And we need the `run`, `issues`, `linters` section too just to be able to use `linters-settings`? 

By the way I have a beef with this rule because at the moment I do not think it is useful. Most of our stuff simply receive this to please the linting gods (you are the exception because I remember in previous cache PRs you left some actual description for the exported functions):
```
// [type-name] ...
```

I am happy to have it as it is too but if it were to me then I would simply remove this rule. 😄

tothszabi (approved):

.golangci.yml
ofalvai: > And we need the run, issues, linters section too just to be able to use linters-settings?

The `run` and `issues` sections are needed to make it work in our repo, the `linters` section is not strictly needed, but I thought it's a good idea to explicitly define the enabled linters that are enabled by default anyway. But I'm happy to remove it if you don't find it useful!

> By the way I have a beef with this rule because at the moment I do not think it is useful. Most of our stuff simply receive this to please the linting gods

I too find it sometimes annoying, but it also made me add meaningful comments on a few occasions when I forgot to do it on my own 😁 Let's go with it for now and I'll raise this on the next eng sync to discuss.

PR #169, opened 2022-10-07
Remove CompareVersions logs

godrei:
This PR removes unnecessary logs from the `version` package.

### DIFF
--- a/versions/versions.go
+++ b/versions/versions.go
@@ -1,7 +1,7 @@
 package versions
 
 import (
-	"log"
+	"fmt"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


versions/versions.go
```diff
@@ -36,14 +36,12 @@ func CompareVersions(version1, version2 string) (int, error) {
...
-			log.Println("Failed to parse int:", err)
-			return -2, err
+			return -2, fmt.Errorf("failed to parse int (%s): %s", num, err)
 		}
 
 		num2, err2 := strconv.ParseInt(version2Slice[i], 0, 64)
 		if err2 != nil {
-			log.Println("Failed to parse int:", err2)
-			return -2, err
+			return -2, fmt.Errorf("failed to parse int (%s): %s", num, err)
```
lpusok, line 44:
```suggestion
			return -2, fmt.Errorf("failed to parse int (%s): %s", version2Slice[i], err2)
```

godrei: Good catch thanks!

lpusok (approved):

PR #172, opened 2022-11-28
Add go exit codes

hisaac:
## Preamble

I'm feeling pretty rusty in my Go skills after returning from parental leave, so I decided to just do a small change here to limber up.

Also, I am not wed to this idea. I think it would be nice, but if the team believes it's superfluous, I'm also happy just closing this. Thanks for taking a look!

## Changes

This change adds a module called `exitcode` that contains a simple enum type called `ExitCode`. This provides a more semantically meaningful exit status in the code.

## More Details

I explain the motivations a little more here: https://github.com/hisaac/GoExitCodes

You can see an example of how this could be used here: https://github.com/bitrise-steplib/steps-fastlane/pull/91


vshah23 (commented):
This might be useful to standardize exit codes across steps/cli if that's something we're interested in.

hisaac:
@vshah23 Agreed! I just noticed today that there was an [exitcode.go](https://github.com/bitrise-io/bitrise/blob/master/exitcode/exitcode.go) file in the CLI now. I think that's a new addition since I went on leave.

ofalvai (approved):
I don't have strong feelings about this, I think we can try this and see if it sticks or not.

godrei (approved):

PR #176, opened 2022-12-29
Add arbitrary options to git clone commands

ofalvai:
### Context

https://github.com/bitrise-io/bitrise/pull/838

### Changes

Add a variadic `opts string...` argument to the `Clone()` functions, matching the other git command functions. This allows passing additional flags to the `git clone` command, such as the `--depth=1` flag that does a shallow clone.

In the case of `CloneTagOrBranch()`, I decided to apply the provided opts after the hardcoded `--recursive` and `--branch` flags, making it possible to override those flags from outside (although I don't think it will be a frequent usage pattern)

### DIFF
--- a/command/git/commands.go
+++ b/command/git/commands.go
@@ -10,13 +10,24 @@ func (g *Git) Init() *command.Model {
 }
 
 // Clone a repository into a new directory.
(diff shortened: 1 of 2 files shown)


hisaac (approved):
Good idea here 👍

PR #178, opened 2023-01-30
rsync output

godrei:
This PR stops rsync printing directly to the standard output.

This is needed because the Bitrise CLI uses `CopyFile` and `CopyDir` functions for activating local Steps.
When rsync fails, it directly prints an error to the stdout, which messes up the CLI's structured logging.

### DIFF
--- a/command/file.go
+++ b/command/file.go
@@ -2,11 +2,27 @@ package command
 
 import (
 	"errors"
+	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ofalvai (approved):
Looks good, but why didn't you modify the original `RunCommand()` method instead? That would solve all other possible logging issues, no?

godrei:
@ofalvai this is the V1 command package, still in use by tools and probably by steps too and I wanted to reduce the blast radius.

vshah23 (approved):

PR #187, opened 2023-11-07
adding a fork of hashicorp's retryable http client which implements m…

rescuerunaways:
…ultithreaded downloader


rescuerunaways:
closing a PR, as we found an _easiser_ solution
may reconsider in the future

PR #193, opened 2024-02-21
Adds UpdateReg git command

lpusok:
Adds UpdateReg git command, it can be used to delete refs.

### DIFF
--- a/command/git/commands.go
+++ b/command/git/commands.go
@@ -47,7 +47,7 @@ func (g *Git) Fetch(opts ...string) *command.Model {
 	return g.command(args...)
 }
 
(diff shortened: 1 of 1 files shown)


command/git/commands.go
```diff
@@ -47,7 +47,7 @@ func (g *Git) Fetch(opts ...string) *command.Model {
 	return g.command(args...)
 }
 
-// Checkout switchs branches or restore working tree files.
+// Checkout switches branches or restore working tree files.
```
ofalvai, line 50: image

ofalvai (approved):

PR #196, opened 2024-05-15
Return ExitError type

lpusok:
Shove the original command error in the error tree. This allows checking for the specific exit code by code like this:
```
	var exitError *exec.ExitError
	if errors.As(err, &exitError) {
```

Added command.FormattedError type to work around issue of not returning the original exec.ExitError type. The Unwrap() method returns the original error, so the above example will now work.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -75,6 +75,22 @@ type Command interface {
 	Wait() error
 }
 
+// CommandError ...
+type CommandError struct {
+	formattedErr       error
+	originalCommandErr error
+}
+
+// Error ...
+func (c *CommandError) Error() string {
+	return c.formattedErr.Error()
+}
+
+// Unwrap ...
+func (c *CommandError) Unwrap() []error {
+	return []error{c.originalCommandErr}
+}
+
 type command struct {
 	cmd            *exec.Cmd
 	errorCollector *errorCollector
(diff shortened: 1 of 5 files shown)


command/command.go
```diff
@@ -75,6 +75,23 @@ type Command interface {
 	Wait() error
 }
 
+// FormattedError ...
+type FormattedError struct {
```
godrei, line 79: I suggest moving the error-related code to a new file (errors.go).

godrei: Can we rename this error to `ExitError` or `ExitStatusError` or `CommandFailedError`, we only use this error type, if the command was executed successfully, but returned a non-zero exit status.

godrei: Can we create a constructor for this error type, which takes care os the actual error formatting, to separate this concept from the command execution logic a bit more:

`func NewCommandFailedError(exitStatus int, prinatbleCmd string, reason error/[]string) ...`

if the reason is empty we could add the placeholder text as a reason (`check the command's output for details`)

command/command_test.go
```diff
@@ -236,6 +259,10 @@ Error: fourth error`,
 				t.Errorf("command.Run() error = %v, wantErr %v", gotErrMsg, tt.wantErr)
 				return
 			}
+			gotFormattedMsg := errorutil.FormattedError(err)
+			if gotFormattedMsg != tt.wantErr {
```
godrei, line 263: gotFormattedMsg shouldn't be the same as gotErrMsg.

gotErrMsg should be (`<error>: <reason>`):

```
command failed with exit status 1 (bash "testdata/exit_with_message.sh"): Error: first error
Error: second error
Error: third error
Error: fourth error
```

while gotFormattedMsg should be `<error>:\n\t<reason>`: 

```
command failed with exit status 1 (bash "testdata/exit_with_message.sh"): 
  Error: first error
  Error: second error
  Error: third error
  Error: fourth error
```

command/command.go
lpusok: Ok, added errors.go.

command/command_test.go
lpusok: Added a new test (only slightly janky) in the errorutil package.

godrei (dismissed):

ofalvai (approved):

PR #197, opened 2024-05-21
Use sterr from exitError in error message as a fallback

lpusok:
Use the [stderr](https://pkg.go.dev/os/exec#ExitError.Stderr) included in exit.ExitError in command error messages as a fallback, if available. This works even we do not provided a command specific error finder (for example tar command).

Example:
```
command failed with exit status 1 (tar "-tf" "/var/folders/r5/gkvczn3j2tb0m79nwby9fjv80000gq/T/Test_compression1308626968/001/compression_test.tzst"): tar: Error opening archive: Child process exited with status 1
```


tothszabi (approved):

ofalvai (approved):
Great idea, thank you!

PR #203, opened 2025-06-02
Handle BITRISE_ANALYTICS_DISABLED in step tracker

andromedakovach:
### Context

This env var is already handled and respected in https://github.com/bitrise-io/bitrise and its tracker implementation, but not in the step-side tracker lib.

### Changes
- add noop tracker implementation
- default tracker checks `BITRISE_ANALYTICS_DISABLED` env var
- tests for disabling analytics

### DIFF
--- a/analytics/track.go
+++ b/analytics/track.go
@@ -2,10 +2,10 @@ package analytics
 
 import (
 	"bytes"
-	"os"
 	"sync"
 	"time"
 
+	"github.com/bitrise-io/go-utils/v2/env"
 	"github.com/bitrise-io/go-utils/v2/log"
 )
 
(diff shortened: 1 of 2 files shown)


analytics/track.go
```diff
@@ -27,8 +30,19 @@ type tracker struct {
...
+
+// Enqueue ...
+func (t noopTracker) Enqueue(eventName string, properties ...Properties) {}
+
+// Wait ...
+func (t noopTracker) Wait() {}
+
 // NewDefaultTracker ...
 func NewDefaultTracker(logger log.Logger, properties ...Properties) Tracker {
+	if disabled, err := strconv.ParseBool(os.Getenv(analyticsDisabledEnv)); err == nil && disabled {
```
godrei, line 43: We could simply check the expected value (when disabled) similarly to the [Bitrise CLI's implementation](https://github.com/bitrise-io/bitrise/blob/master/analytics/state_checker.go#L26):
`s.envRepository.Get(DisabledEnvKey) != "true"`

This would simplify the code and also unify the behaviour amond the two components (I'm not sure, but `strconv.ParseBool` might consider "yes" as a true boolean value, while the CLI only accepts "true").

godrei (commented):
Once we merge this PR, we can remove the same check from the [Bitrise CLI's analytics implementation](https://github.com/bitrise-io/bitrise/blob/master/analytics/tracker.go#L134), to avoid double checking the same config.

godrei (approved):

PR #205, opened 2025-08-25
feat: Allow overriding log output

zsolt-marta-bitrise:


log/log.go
```diff
@@ -27,15 +27,48 @@ type Logger interface {
...
 type logger struct {
 	enableDebugLog  bool
 	timestampLayout string
 	stdout          io.Writer
 }
 
 // NewLogger ...
-func NewLogger() Logger {
-	return &logger{enableDebugLog: false, timestampLayout: defaultTimeStampLayout, stdout: os.Stdout}
+func NewLogger(options ...LoggerOptions) Logger {
```
godrei, line 40: Q: Why not passing an `opts` struct as a variadic parameter, which stores the logger config?

```
type Opts structure {
	EnableDebugLog  bool
	TimestampLayout string
	Stdout          io.Writer
}
```

zsolt-marta-bitrise: Because this way we don't have to set all the options, just the ones we want to overwrite

godrei (approved):

log/log.go
zsolt-marta-bitrise: Yeah I really suggest using this pattern, it's a bit more verbose, but CoPilot is great at generating it :D

PR #229, opened 2026-04-23
[STEP-2150] Migrate envutil helpers to env.Repository + deprecated compat package

Badlazzor:
## Summary

Brings the v1 \`envutil\` helpers into v2 as a **hybrid**:

1. \`env.Repository\` gains helper methods so new callers get a testable, mockable API on the interface that already owns env state.
2. A new \`envutil\` package carries the v1 function names as thin package-level wrappers around \`env.NewRepository()\`, each marked \`// Deprecated:\` and pointing at the Repository method. Existing callers migrate by import-path rename only.

### New \`env.Repository\` methods

- \`GetOrDefault(key, def string) string\`
- \`Required(key string) (string, error)\`
- \`FlagOrEnv(flag *string, key string) string\`
- \`Revokable(key, value string) (func() error, error)\`
- \`RevokableMany(envs map[string]string) (func() error, error)\`
- \`Scoped(key, value string, fn func()) error\`
- \`ScopedMany(envs map[string]string, fn func()) error\`

\`Scoped\` / \`ScopedMany\` restore the previous values via \`defer\` so they work even when \`fn\` panics.

### \`envutil\` package — v1-compat (Deprecated)

| v1 name | Delegates to |
|---------|--------------|
| \`RevokableSetenv\` | \`env.Repository.Revokable\` |
| \`RevokableSetenvs\` | \`env.Repository.RevokableMany\` |
| \`SetenvForFunction\` | \`env.Repository.Scoped\` |
| \`SetenvsForFunction\` | \`env.Repository.ScopedMany\` |
| \`StringFlagOrEnv\` | \`env.Repository.FlagOrEnv\` |
| \`GetenvWithDefault\` | \`env.Repository.GetOrDefault\` |
| \`RequiredEnv\` | \`env.Repository.Required\` |

### Notes

- \`env.Repository\` is an interface and has grown new methods. **Downstream mocks must regenerate.** v2 is in alpha, that cost is expected.
- Known callers (per code search): \`bitrise-plugins-io\`, \`bitrise-api\`, \`bitrise-addon-service\`, \`bitrise-devops-microservice-common\`, \`den\`, \`support-utils\`, \`gotgen\`, \`bitrise-plugins-service\`. No direct step callers — envutil was only used in backend services and tools.

## Test plan

- [x] \`go test ./env/... ./envutil/...\` — all cases pass (including scoped-panic restoration)
- [x] \`go test ./...\` — repo-wide tests pass
- [x] \`go vet ./...\` clean
- [ ] End-to-end validation with one caller (follow-up draft PR pinning \`go-utils/v2\` to this branch)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## End-to-end validation

Intentionally **no consumer PR** for this ticket:

- No step callers. Every known v1 envutil caller is a backend service or legacy tool (`bitrise-plugins-io`, `bitrise-api`, `bitrise-addon-service`, `bitrise-devops-microservice-common`, `den`, `support-utils`, `gotgen`, `bitrise-plugins-service`).
- Every one of those callers is on a legacy module state (Go 1.12–1.18, some with no `go` directive), so adding `go-utils/v2` would force a toolchain bump + full vendor regen that dwarfs the actual API change.

Parent ticket STEP-2134 AC "Test with a Step" does not apply here — envutil has no step consumers. Callers can validate when they migrate on their own timeline.

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -141,33 +141,36 @@ func RevokableMany(r GetSetter, envs map[string]string) (func() error, error) {
 }
 
 // Scoped sets key to value on r, invokes fn, then restores the previous value.
-// The restore runs even if fn panics.
+// The restore runs even if fn panics, and even when the initial Set reports
+// an error (a GetSetter may leave state written and still return an error).
 func Scoped(r GetSetter, key, value string, fn func()) (err error) {
 	revoke, setErr := Revokable(r, key, value)
-	if setErr != nil {
-		return setErr
-	}
 	defer func() {
 		if rerr := revoke(); err == nil {
 			err = rerr
 		}
 	}()
+	if setErr != nil {
+		return setErr
+	}
 	fn()
 	return nil
 }
 
 // ScopedMany applies every entry in envs on r, invokes fn, then restores all
-// previous values. Restore runs even if fn panics.
+// previous values. Restore runs even if fn panics. On setup failure
+// RevokableMany has already restored every key it touched and returns a
+// no-op revoke, so the deferred call is a safe cleanup either way.
 func ScopedMany(r GetSetter, envs map[string]string, fn func()) (err error) {
 	revoke, setErr := RevokableMany(r, envs)
-	if setErr != nil {
-		return setErr
-	}
 	defer func() {
 		if rerr := revoke(); err == nil {
 			err = rerr
 		}
 	}()
+	if setErr != nil {
+		return setErr
+	}
 	fn()
 	return nil
 }
--- a/env/env.go
+++ b/env/env.go
@@ -1,6 +1,7 @@
 package env
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 )
@@ -22,12 +23,43 @@ func (l commandLocator) LookPath(file string) (string, error) {
 	return exec.LookPath(file)
 }
 
-// Repository ...
+// Repository abstracts read/write access to process environment variables.
+// Implementations should be safe to replace in tests (e.g. with an
+// in-memory fake) without touching the real process environment.
 type Repository interface {
+	// List returns the current environment as "KEY=VALUE" entries, like
+	// os.Environ().
 	List() []string
+	// Unset removes key from the environment.
 	Unset(key string) error
+	// Get returns the value of key, or "" when unset.
 	Get(key string) string
+	// Set assigns value to key.
 	Set(key, value string) error
+
+	// GetOrDefault returns the value of key or def when the key is unset
+	// or empty.
+	GetOrDefault(key, def string) string
+	// Required returns the value of key or an error when it is unset or
+	// empty.
+	Required(key string) (string, error)
+	// FlagOrEnv returns *flag when it points to a non-empty string, else
+	// the value of key. A nil pointer is treated as unset.
+	FlagOrEnv(flag *string, key string) string
+
+	// Revokable sets key to value and returns a function that restores
+	// the previous value when invoked.
+	Revokable(key, value string) (revoke func() error, err error)
+	// RevokableMany applies every entry in envs and returns a single
+	// function that restores all previous values. If any Set fails,
+	// the returned revoke still restores entries already applied.
+	RevokableMany(envs map[string]string) (revoke func() error, err error)
+	// Scoped sets key to value, invokes fn, and then restores the
+	// previous value. The restore runs even if fn panics.
+	Scoped(key, value string, fn func()) error
+	// ScopedMany applies every entry in envs, invokes fn, and restores
+	// all previous values. Restore runs even if fn panics.
+	ScopedMany(envs map[string]string, fn func()) error
 }
 
 // NewRepository ...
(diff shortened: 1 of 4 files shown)


env/env.go
```diff
@@ -22,12 +23,43 @@ func (l commandLocator) LookPath(file string) (string, error) {
...
+	// os.Environ().
 	List() []string
+	// Unset removes key from the environment.
 	Unset(key string) error
+	// Get returns the value of key, or "" when unset.
 	Get(key string) string
+	// Set assigns value to key.
 	Set(key, value string) error
+
+	// GetOrDefault returns the value of key or def when the key is unset
```
benbitrise: what about [interface at the call site](https://bitrise.atlassian.net/wiki/spaces/STEP/pages/4934795283/Go+Interface+Patterns+Late+Definition+Minimal+Mocks)?

env/env.go
```diff
@@ -56,3 +88,85 @@ func (d repository) Unset(key string) error {
...
+				return err
+			}
+		}
+		return nil
+	}
+
+	for k, v := range envs {
+		originals[k] = d.Get(k)
+		if err := d.Set(k, v); err != nil {
+			return revoke, err
```
benbitrise: The way I read this is....


if i provide 10 envs and the first 5 succeed, but the 6th has an error, the loop exits early without items 7+ getting set. And the caller doesn't know which were set and which were not. Do we want this behavior?

env/env.go
```diff
@@ -56,3 +88,85 @@ func (d repository) Unset(key string) error {
...
+	return revoke, nil
+}
+
+// Scoped ...
+func (d repository) Scoped(key, value string, fn func()) (err error) {
+	revoke, setErr := d.Revokable(key, value)
+	if setErr != nil {
+		return setErr
+	}
+	defer func() {
```
benbitrise: Question: suppost someone defines a `repository` that successfully sets the variable in `Revokable`, but for some other reason reports an error, then the defer with `revoke()` is never called. Should the defer get moved up?

If I am being too imaginative with how things can wrong, then just let me know and we can skip past this comment.

envutil/envutil.go
```diff
@@ -0,0 +1,71 @@
+// Package envutil exposes the v1 envutil package-level function names on
```
benbitrise, line 1: I don't understand how this approach helps us meet our goals of v1 deprecation. It seems like it just prolongs use of a pattern we are trying to move away from.

env/env.go
Badlazzor: I was bringing it over verbatim, but you are right, I shouldn't go back on my words. Thanks for keeping me honest.

envutil/envutil.go
Badlazzor: It migrates the lib, and allows us to do the migration in dependants after we merge this. We can of course go the way I've done with `output` but that raised concerns and got pushback. This allows finishing these tasks and do the usage site updates in the following tickets.

env/env.go
Badlazzor: Good catch. Inherited from v1 (ad86bf7) verbatim but contract was weak: caller received a `revoke` covering only keys written so far and no signal about which subset succeeded — and since Go map iteration is randomized, "first 5" differs per run anyway.

Switched `RevokableMany` to atomic-on-error in 67a5225: on any `Set` failure we restore every key already written, return a no-op revoke, and wrap the underlying error (plus any restore failure). Net env state on error == pre-call state. `ScopedMany` inherits the new semantics for free.

Added a regression test that uses an empty key (`os.Setenv` rejects it) mixed with valid ones so the assertion holds regardless of iteration order.

env/env.go
Badlazzor: Taken care of in 86788b1. `Repository` interface is back to its pre-existing shape (List/Unset/Get/Set), no downstream type-signature churn. The new helpers moved to package-level funcs taking minimal call-site interfaces:

```go
type Getter interface{ Get(key string) string }
type GetSetter interface {
    Get(key string) string
    Set(key, value string) error
}

func GetOrDefault(r Getter, key, def string) string { ... }
func Revokable(r GetSetter, key, value string) (func() error, error) { ... }
// etc.
```

`envutil` shims now delegate to `env.Revokable`, `env.Scoped`, etc. with the package's default repo — public surface of the shims is unchanged.

env/env.go
Badlazzor: Not too imaginative — valid concern. A custom `GetSetter` is free to write state and still return an error, and the early return dropped the restore.

Fixed in 19afb45 by moving the defer above the setErr check in both `Scoped` and `ScopedMany`. `revoke` is always non-nil: `Revokable` constructs it before invoking `Set`, and `RevokableMany` returns a no-op revoke on error (it already restored every key it touched itself), so the deferred call is safe on every path.

Regression test uses a fake GetSetter that writes-then-errors. Without the fix it would leave the store at the "new" value; with the fix the store ends up back at the original.

benbitrise (approved):

PR #230, opened 2026-05-22
Clarify ziputil behavior with comments and tests

godrei:
Replace placeholder `...` comments with accurate documentation for all public functions. Document non-obvious behaviors: `isContentOnly` archive structure, `ZipDirs` rsync-based merge semantics and `log.Fatal` on cleanup failure, `ZipFiles` no-output guarantee on duplicate basenames, and `UnZip` zip-slip strip-and-continue behavior. Tighten inline flag comments (`-r`, `-T`, `-y`, `-j`) to describe what each flag actually does.

Rewrite the test file: remove weak tests that only assert the output file exists (all covered by existing round-trip tests), strengthen two unique scenarios into proper round-trips with content checks, fix a `os.MkdirAll` copy-paste bug, and add four new tests that lock in the behaviors above: directory mtime preservation, same-basename directory merge, no output file on duplicate basename error, and zip-slip strip-and-extract.

### DIFF
--- a/ziputil/ziputil.go
+++ b/ziputil/ziputil.go
@@ -34,9 +34,9 @@ func ZipDir(sourceDirPth, destinationZipPth string, isContentOnly bool) error {
 }
 
 // ZipDirs zips multiple directories into a single archive, each under its own basename.
-// When two entries in sourceDirPths share the same basename (e.g. "/a/shared" and "/b/shared"),
-// their contents are merged via rsync (-ar); the outcome for conflicting filenames depends
-// on rsync's overwrite rules and the files' mtimes at the time of the call.
+// When two entries in sourceDirPths share the same basename (e.g. "/a/shared" and
+// "/b/shared"), their contents are merged; files unique to either are preserved,
+// and conflicting filenames are overwritten by entries later in sourceDirPths.
 // If the temporary directory cleanup fails, the process is terminated via log.Fatal.
 func ZipDirs(sourceDirPths []string, destinationZipPth string) error {
 	for _, path := range sourceDirPths {
@@ -70,7 +70,8 @@ func ZipDirs(sourceDirPths []string, destinationZipPth string) error {
 
 func internalZipDir(destinationZipPth, zipTarget, workDir string) error {
 	// -r: recurse into directories
-	// -T: test archive integrity after creation; deletes the archive on failure
+	// -T: test the temp archive with unzip before overwriting the destination;
+	//     on failure the destination is left in its prior state
 	// -y: store symlinks as symlinks instead of following them
 	cmd := command.New("/usr/bin/zip", "-rTy", destinationZipPth, zipTarget)
 	cmd.SetDir(workDir)
@@ -99,7 +100,8 @@ func ZipFiles(sourceFilePths []string, destinationZipPth string) error {
 		}
 	}
 
-	// -T: test archive integrity after creation; deletes the archive on failure
+	// -T: test the temp archive with unzip before overwriting the destination;
+	//     on failure the destination is left in its prior state
 	// -y: store symlinks as symlinks instead of following them
 	// -j: junk directory paths, store files under their base names only
 	parameters := []string{"-Tyj", destinationZipPth}
--- a/ziputil/ziputil.go
+++ b/ziputil/ziputil.go
@@ -10,7 +10,10 @@ import (
 	"github.com/bitrise-io/go-utils/pathutil"
 )
 
-// ZipDir ...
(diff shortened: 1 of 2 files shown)


ziputil/ziputil.go
```diff
@@ -62,9 +69,9 @@ func ZipDirs(sourceDirPths []string, destinationZipPth string) error {
 }
 
 func internalZipDir(destinationZipPth, zipTarget, workDir string) error {
-	// -r - Travel the directory structure recursively
-	// -T - Test the integrity of the new zip file
-	// -y - Store symbolic links as such in the zip archive, instead of compressing and storing the file referred to by the link
+	// -r: recurse into directories
+	// -T: test archive integrity after creation; deletes the archive on failure
```
Badlazzor: `-T` does not "delete the archive on failure". `zip -h2` describes it as: *test completed temp archive with unzip before updating archive*. zip writes a temp archive, tests it, then renames into place; on test failure the **temp** is removed and the destination is left in its prior state (untouched if pre-existing, never created otherwise).

Suggested:
```go
// -T: test the temp archive with unzip before overwriting the destination;
//     on test failure the destination is left in its prior state.
```

Same fix needed at line 102.

ziputil/ziputil.go
```diff
@@ -30,7 +33,11 @@ func ZipDir(sourceDirPth, destinationZipPth string, isContentOnly bool) error {
 
 }
 
-// ZipDirs ...
+// ZipDirs zips multiple directories into a single archive, each under its own basename.
+// When two entries in sourceDirPths share the same basename (e.g. "/a/shared" and "/b/shared"),
+// their contents are merged via rsync (-ar); the outcome for conflicting filenames depends
+// on rsync's overwrite rules and the files' mtimes at the time of the call.
```
Badlazzor: This doc contradicts the new `TestZipDirsSameBasenamesMerge`, which asserts a deterministic last-wins outcome (`"second"`).

`command.CopyDir` calls `rsync -ar` without `--update`, so rsync overwrites unconditionally — not mtime-dependent. The behavior is deterministic: entries later in `sourceDirPths` overwrite conflicting files from earlier entries.

Suggested:
```go
// When two entries in sourceDirPths share the same basename (e.g. "/a/shared" and
// "/b/shared"), their contents are merged; files unique to either are preserved,
// and conflicting filenames are overwritten by entries later in sourceDirPths.
```

ziputil/ziputil_test.go
```diff
@@ -399,6 +270,138 @@ func TestUnZipDirectory(t *testing.T) {
...
+
+	// Despite the error, the file is extracted inside destDir with the traversal stripped.
+	exist, statErr := pathutil.IsPathExists(filepath.Join(destDir, "escape.txt"))
+	require.NoError(t, statErr)
+	require.True(t, exist, "system unzip strips '../' and extracts inside dest")
+
+	// The file must not have escaped outside destDir.
+	escaped, statErr := pathutil.IsPathExists(filepath.Join(tmpDir, "escape.txt"))
+	require.NoError(t, statErr)
+	require.False(t, escaped, "file must not escape outside intoDir")
```
Badlazzor, line 409: Negative assertion checks the wrong path. `destDir = tmpDir/dest`, so a `../../escape.txt` entry from inside `destDir` resolves to `parentOf(tmpDir)/escape.txt`, not `tmpDir/escape.txt`. This check passes regardless of whether a real escape happened.

Suggested:
```go
escapeTarget := filepath.Join(filepath.Dir(tmpDir), "escape.txt")
escaped, statErr := pathutil.IsPathExists(escapeTarget)
require.NoError(t, statErr)
require.False(t, escaped, "file must not escape outside intoDir")
```

Separately: zip-slip behavior is `unzip` implementation specific (Info-ZIP vs others). Test may behave differently on Linux distros with non–Info-ZIP unzip — worth a `runtime.GOOS` guard or a comment documenting the assumption.

Badlazzor (commented):
Three accuracy nits before merge.

ziputil/ziputil.go
godrei: fixed in https://github.com/bitrise-io/go-utils/pull/230/commits/1447e8b8fb6a3b770aa456508fb57356c1d3c7b0

ziputil/ziputil.go
godrei: fixed in https://github.com/bitrise-io/go-utils/commit/1447e8b8fb6a3b770aa456508fb57356c1d3c7b0

Badlazzor (approved):

PR #231, opened 2026-05-22
Ziputil migration

godrei:
Migrates the `ziputil` package from v1 (shells out to `/usr/bin/zip` and `/usr/bin/unzip`) to a pure Go stdlib implementation backed by `archive/zip`.

PR [230](https://github.com/bitrise-io/go-utils/pull/230) documents the v1 behavior in comments and tests to establish a clear behavioral contract before migration.

## What changed

- `ZipDir`, `ZipDirs`, `ZipFile`, `ZipFiles`: pure Go replacements for `/usr/bin/zip -rTy` / `-Tyj`. Directory modification times are preserved; symlinks are stored as symlinks (not followed).
- `UnZip`: pure Go replacement for `/usr/bin/unzip`. Restores file permissions and symlinks. Creates the destination directory if it does not exist (consistent with `unzip -d` behavior).
- `ZipManager` struct with `OsProxy` interface for testability without the real OS.
- On a write failure, the destination is left in its prior state, same as v1 (v2 writes to a temp file and renames over the destination only on success).

## Behavioral differences from v1

**`ZipFiles` duplicate base names:** v2 detects duplicates before creating the destination file and returns an error immediately, so no partial output is left behind. v1 relied on `/usr/bin/zip` to fail after the fact.

**`ZipFiles` symlinks:** both v1 (`zip -y`) and v2 store symlinks as symlinks rather than following them. v2 achieves this by detecting `ModeSymlink` via `Lstat` and routing to the symlink-storage path.

**`UnZip` path-traversal (zip-slip):** v1 used the system `unzip`, which stripped `../` components, extracted the file to the sanitized path inside the destination directory, and returned a non-zero exit code. v2 rejects the first traversal entry and returns an error immediately; no further entries are extracted after detection. This is strictly safer.

**`UnZip` chained-symlink attack:** v2 resolves pre-existing symlinks in the parent directory via `EvalSymlinks` before placing each extracted file or symlink. This prevents a crafted archive from using a previously-extracted symlink to redirect an entry outside the destination. v1 had no such protection.

**`ZipDirs` temp-dir cleanup:** v1 called `log.Fatal` if the temporary directory cleanup failed. v2 has no temporary directory.

## Tests

Every test scenario from the v1 clarification (PR 230) is carried over to v2 (some as unit tests, some as integration tests), adapted to the `ZipManager` API and to v2's behavioral differences (e.g. zip-slip is asserted as a hard rejection rather than strip-and-continue). This keeps the documented v1 behavioral contract under test through the migration.

Unit tests cover all public functions. Integration tests (tagged `//go:build integration`) verify content correctness, directory mtime preservation, same-basename merge semantics, the no-output guarantee on duplicate basenames, and zip-slip rejection.


Badlazzor (commented):
Pulled the branch, ran unit + integration tests on Go 1.26.3 — all green. The behavioral changes vs v1 are well-documented and the security posture (strict zip-slip reject, chained-symlink defense, deterministic merge semantics) is a solid upgrade. A few findings, mostly tests + small consistency nits — none blocking, listed roughly by impact:

**1. Test gap on the security-critical paths.** The PR body calls out the chained-symlink defense and the symlink target validation as the main hardening over v1, but no test currently exercises them:

- `extractEntry` `EvalSymlinks` parent check (ziputil.go:324–330 for symlinks, :342–348 for regular files): a pre-existing symlink in the parent path should cause extraction to fail.
- `filepath.IsAbs(targetStr)` reject (ziputil.go:311–313).
- `resolvedTarget` escape via relative `../../` symlink target (ziputil.go:315).

These are the most attractive attack surfaces in an unzipper. Worth adding integration tests that craft archives with: (a) a benign first entry that creates `evil -> /tmp/escape`, then a second entry that writes through `evil/foo`; (b) a symlink entry with an absolute target; (c) a symlink entry with `../../etc/foo` as target.

**2. `OsProxy` interface is defined but never used for mocking.** Every test calls `NewZipManager(pathutil.NewPathChecker())` against the real OS, so the interface adds API surface with no current consumer. Two reasonable directions:
- Keep it and add mock-driven tests for OS error paths (e.g., `Create` failure leaves no destination file; `Symlink` failure surfaces cleanly).
- Drop it and revert to direct `os.*` calls until a real second implementation exists.

I'd lean toward the second if there's no near-term plan to test OS errors — the current shape is interface-without-purpose.

**3. Inconsistent abstraction in cleanup paths.** `ZipDirs`, `ZipFiles`, and `createZipFromDir` all clean up partial output via raw `os.Remove` (ziputil.go:65, :115, :181) rather than `z.osProxy`. If `osProxy` is the OS boundary, this leaks — and prevents a mock from observing cleanup behavior on failure. Easy to route through `osProxy.Remove` (would need adding to the interface).

**4. Stale comment on `extractEntry` (ziputil.go:288–292).** "Belt-and-suspenders: UnZip pre-filters with strings.Contains before calling here, but this guard protects against any future direct callers." `extractEntry` is unexported, so the only possible "future direct callers" are within the same package — which is a much weaker argument. The real reason looks like defensive CodeQL-style hardening; worth saying that explicitly so future readers don't try to "simplify" by dropping the guard.

**5. Tradeoff documented in the PR body isn't in godoc.** The `strings.Contains(f.Name, "..")` guard (ziputil.go:163) rejects legitimate names like `module..framework`. PR body explicitly accepts the tradeoff for iOS/Xcode artifacts, but a user encountering the rejection later won't see that context. Worth a sentence in `UnZip`'s godoc:

> Entries whose name contains the substring `..` anywhere (including names like `foo..bar`) are rejected. This is broader than strict path-component traversal but is required for the static analysis used in this repo to recognise the sanitizer.

**6. Tests are flat.** Most fixtures are one-level directories with a handful of files. Worth one integration test with a deeper mix — nested subdirs, a symlink inside a subdir, an empty subdir — to stress `addDirToZip` + `WalkDir` + per-entry routing together. Right now the symlink + directory paths are tested, but not their combination.

Happy to send a follow-up PR with the test cases in #1 if useful.

godrei:
@Badlazzor 
- Security-check tests added in https://github.com/bitrise-io/go-utils/pull/231/commits/14555e716893670f44a2d888e71c2d8bbc468995
- OsProxy mocking introduced in https://github.com/bitrise-io/go-utils/pull/231/commits/30253317847deb956ec113540fa7b5302b6d461e
- Raw os.Remove cleanup in https://github.com/bitrise-io/go-utils/pull/231/commits/773c963c65c3383578fb60a47e23d090e979610f
- Stale extractEntry comment fixed and .. godoc tradeoff documented in https://github.com/bitrise-io/go-utils/pull/231/commits/ff8b5ae393c266d8de660772c06b22517b522047
- Deep-tree test added in https://github.com/bitrise-io/go-utils/pull/231/commits/fa80291d08716897f82db3088e0f4b7abe093d17

Badlazzor (dismissed):

marcell-vida (approved):

PR #234, opened 2026-08-27
feat(v1->v2): `Kill` function for command

marcell-vida:
`Command` exposes `Start()` and `Wait()` but keeps its `*exec.Cmd` private, so a running command cannot be stopped.

v1 allowed this previously: https://github.com/bitrise-io/go-utils/blob/v1/command/command.go#L56

### Notable changes:

- `Signal` and `Kill` functions
- Breaking: hand-rolled `Command` fakes need a stub
- `ErrProcessFinished` and `ErrProcessNotStarted` errors
- Signalling after `Wait()` cannot reach a recycled PID
- `mocks/Command.go` hand-edited, `command` absent from `.mockery.yml`

`Opts.ProcessGroup` was prototyped and dropped, no orphans on Linux.

### Consumer

Main motivator was the AVD Manager migration: https://github.com/bitrise-steplib/steps-avd-manager/pull/70


command/command.go
```diff
@@ -152,6 +163,30 @@ func (c command) Wait() error {
...
+func (c command) Signal(sig os.Signal) error {
+	if c.cmd.Process == nil {
+		return ErrProcessNotStarted
+	}
+	// os.Process tracks completion itself, so signalling after Wait reports ErrProcessDone
+	// instead of reaching a process that reused the PID. Reading cmd.ProcessState here would
+	// race with Wait.
+	err := c.cmd.Process.Signal(sig)
+	if errors.Is(err, os.ErrProcessDone) || errors.Is(err, syscall.ESRCH) {
+		return ErrProcessFinished
```
lpusok: Can we return os.ErrProcessDone here? or cannot due to syscall.ESRCH

command/command.go
```diff
@@ -152,6 +163,30 @@ func (c command) Wait() error {
...
+	if err != nil {
+		return fmt.Errorf("signalling command failed (%s): %w", c.PrintableCommandArgs(), err)
+	}
+
+	return nil
+}
+
+// Kill ...
+func (c command) Kill() error {
+	return c.Signal(os.Kill)
```
lpusok, line 178: Should we handle ErrProcessFinished here or leave to the consumer of this package? Ignoring it here would allow to prevent the possible race when we Kill a process, but it exits on its own, with no custom handling needed.

command/command.go
marcell-vida: Good catch, actually Signal already maps that to `ErrProcessDone`. I took out the branch, and the remaining one wraps any errors (including `ErrProcessDone`).

command/command.go
marcell-vida: I don't have a strong opinion on this. If we were to keep it in symmetry with the `Signal` function and return the error, users  could utilize the information or treat it as a non-error later.

lpusok: Wew can go with leaving handling to consumer, it is not heavily used anyway.

lpusok (approved):

